PR #8, opened 2017-08-15
Make Device Access compile on macOS

smarsching:
In the original version the Device Access library does not compile on macOS. This is a pity, because the Control System Adapter depends on this library and there are a number of scenarios where using the code makes sense on non-Linux systems.

The problems that appear when compiling on macOS can be put in two categories: problems caused by using LLVM instead of GCC and problems caused by different APIs (BSD vs Linux). This PR addresses both kind of issues.

In order to make the build work on macOS, the call to the Linux-specific `get_current_dir_name` function has been replaced with a call to the `getcwd` function, which works on both platforms. The PCIe backend and tests that depend on it are only built on Linux because there is no reasonable way how it could be ported to other platforms. Finally, a few linker flags have been changed.

I have verified that with the changes the code still builds on Linux (Ubuntu 16.04 LTS).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -116,7 +116,7 @@ set_target_properties(${PROJECT_NAME} PROPERTIES VERSION ${${PROJECT_NAME}_FULL_
 
 link_directories(${CMAKE_LIBRARY_OUTPUT_DIRECTORY})
 link_directories(${Boost_LIBRARY_DIRS})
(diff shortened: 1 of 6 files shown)


killenb:
Thanks for the pull request. It solves quite some problems which were on our agenda since some time, but with low priority.

* LLVM should have worked. We tried it under Linux with CLANG and only had issues with the coverage report (when dynamically unloading backends it could not find the according source code).
 * We originally had tried to replace get_current_dir_name with a boost filesystem call to make it platform independent, but had to revert it because boost had issues if the language setting were not en_US.UTF-8. 

I will try to get a Mac into our Jenkins configuration so we are sure we do not break it again.

PR #13, opened 2018-03-17
Move PCIe backend to separate directory

smarsching:
By moving the source files for the PCIe backend to a separate directory, we can ensure that it is only built on platforms where it is actually supported. At the moment, this is only Linux (because of dependencies on Linux-specific header files), but this list can easily be extended in the future.

In particular, this change needs to be implemented to make DeviceAccess build on macOS.


smarsching:
I just pushed a second commit that fixed another problem on macOS: The automatically generated `mtca4u-deviceaccess-config` script included on wrong linker flags. This did not cause any problems when building DeviceAccess itself, but it would cause problems when building an application that linked against DeviceAccess and used that script for determining the linker flags.

BTW: The control-system adapter had the same problem. There, I directly pushed my fix to the master branch.

PR #23, opened 2019-01-31
Fix build problems on macOS.

smarsching:
This PR makes two changes:

1. It avoids the use of deprecated / removed C++ features. This is an easy fix because those features where only needed before lambda expressions where added to C++.
2. It defines the `_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR` preprocessor macro so that `std::auto_ptr` is made available when compiling in C++ 17 mode and with libc++.

The second change is not very elegant, but needed because libxml++ 2.6 needs `std::auto_ptr`. This has been fixed in more recent releases of libxml++, so we should check whether we can move to libxml++ 3.x as soon as practical.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -16,7 +16,7 @@ endif()
 include(cmake/enable_latest_cxx_support.cmake)
 
 # C++ compiler flags needed to compile this project and against this project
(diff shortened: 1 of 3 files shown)


mhier:
Thanks for this improvement/fix. Can you please add a comment that _LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR is only required for libxml++2.6?

smarsching:
Done. BTW: Is there any particular reason why we are still on libxml++ 2.6? I think that the API changed slightly in version 3, but if there is no other reason, I would have a look at whether we can move to that new API.

PR #31, opened 2019-02-19
Change ReadAnyGroup::waitAny() to allow deferred processing

smarsching:
The `waitAny` method now returns a notification object instead of a `TransferElementID`. This way, the actual reading of the data update can be deferred until later.

As a side effect, this makes the code cleaner, because now it is not necessary any longer to call `ReadAnyGroup::postRead()`. Instead, the notification object has an `accept()` method that takes care of this. As this method is also called from the destructor of the notification object, it is no longer possible to "forget" to call `postRead()`.

@mhier: This is the change that we discussed offline. As a side effect, polled elements are now only processed when using `readAny()`. Code using `waitAny()` directly has to also call `processPolled()`. This is cleaner than polling these elements from `postRead()` (which was an undocumented side-effect in the old code).

I changed the test code to account for this change and the `readAsync` test runs fine for me now. Some other tests fail, but they also fail on the `master` branch (I guess this is a macOS vs. Linux issue).

Another side effect of this change is that spurious update notifications (notifications that are sent for updates that are dropeed because of an old version number) are no visible to the user (when using `waitAny` instead of `readAny`). I make this information available through the return code of the `accept()` method, which in my opinion is cleaner than having the `detail::DiscardValueException` bubble up.

### DIFF
--- a/device/include/ReadAnyGroup.h
+++ b/device/include/ReadAnyGroup.h
@@ -17,6 +17,81 @@ namespace ChimeraTK {
   /** Group several registers (= TransferElement) to allow waiting for an update of any of the registers. */
   class ReadAnyGroup {
    public:
+    /** Notification object returned by waitAny(). A notification can be accepted immediately or retained to be
+     *  accepted at a later point in time. */
+    class Notification {
+     public:
+      friend class ReadAnyGroup;
+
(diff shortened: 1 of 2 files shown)


mhier:
This looks good, thanks for your help!

PR #38, opened 2019-03-04
added dlmopen as first choice to load the plugin.

nshehz:
Please review. If approve, please merge the branch and close the ticket.


mhier:
testRunTimeBackendLoading crashes in a very strange way for me, I don't really understand what's going wrong there.

mhier:
Unrelated to my previous comment: instead of the assert(backend_name != nullptr) you should throw a proper exception when not all required symbols are found.

nshehz:
The best I can see at the moment with gdb that it crashes as it tries to release memory in malloc.c 
if (chunk_is_mmapped (p))                       /* release mmapped memory. */
{
/* See if the dynamic brk/mmap threshold needs adjusting.
Dumped fake mmapped chunks do not affect the threshold.  */
if (!mp_.no_dyn_threshold
&& chunksize_nomask (p) > mp_.mmap_threshold
&& chunksize_nomask (p) <= DEFAULT_MMAP_THRESHOLD_MAX
&& !DUMPED_MAIN_ARENA_CHUNK (p))
{
......

mhier:
It was decided not to merge, since the solution with dlmopen does not work

PR #156, opened 2020-06-08
Transfer element v1.0 rc5

killenb:
The implementation of the TransferElement showed several issues with the specification in the combination of exception handling and decorators
- B.6.1 was not possible with decorators
- Meta data could not be handled correctly by the ApplicationCore::ExceptionHandlingDecorator.

This behaviour change requires two reviewers to sign off the changes.


mhier:
C.2.2.1: We could implement this in the base class in postXxx() right after doPostXxx() is called. This would help backend implementations. For decorators, we will put the necessary functionality into the decorator base class, so the individual decorators do not have to bother so much (just set the flag before throwing).

mhier:
C.4.2.1: I think this is incorrect. The transfer internally was successful, but the data is rejected. This is similar to the rejection of data in bidirectional variables when the VersionNumber is too old: the transfer technically was ok, but it is discarded and hence it is not a valid transfer. We throw an exception after all, so there is no return value indicating there was new data. We also do not update the data buffer (we cannot, because the values are out of range). Hence we should also not touch the VersionNumber and the DataValidity.

killenb:
> C.2.2.1: We could implement this in the base class in postXxx() right after doPostXxx() is called. This would help backend implementations. For decorators, we will put the necessary functionality into the decorator base class, so the individual decorators do not have to bother so much (just set the flag before throwing).

I was thinking the same. I will change the according section in the spec.

killenb:
> C.4.2.1 (was not changed here, but anyway): I think this is incorrect. The transfer internally was successful, but the data is rejected. This is similar to the rejection of data in bidirectional variables when the VersionNumber is too old: the transfer technically was ok, but it is discarded and hence it is not a valid transfer. We throw an exception after all, so there is no return value indicating there was new data. We also do not update the data buffer (we cannot, because the values are out of range). Hence we should also not touch the VersionNumber and the DataValidity.

Obviously one would have to update the dataValidity to faulty. After catching the exception the TE would then be in a state as the exception handling decorator would leave it, but you can still correlate the version number in data consistency groups. You would lost this information if you don't update the version number.

However, it is some special behaviour in one particular case which is different from all the other exceptions. From the DeviceAccess point of view it is much more consistent if the user application buffer (all of its components) is  never changes when an exception is thrown. Should we put it this explicit if we have not done so already?

killenb:
The requested changes are in from c794f2566b731

PR #158, opened 2020-06-12
[wip] TransferElement spec: Clarified C.2.3

killenb:
I noticed that we have to be more precise in B.2.3 (decorators have to swap the active exception with the target).
As all decorators have to implement it, we should be explicit here.


mhier:
1. "If and only if _activeException == true" is wrong, should be "If and only if _activeException != nullptr". It is not a boolean but a std::exception_ptr.

2. "Decorators must not swap a nullptr into the target because you could swap an active exception out of it." - What ensures that the outer layer does not get an exception swapped out when swapping in a real exception?

3. "This would mean it is thrown too late (only after the decorator has left its doPostXxx(), not inside doPostXxx() when delegating to target->postXxx()), and code that should be skipped could be executed." - This is IMO worded a bit confusing. I would write something like "This would mean it is thrown on the wrong layer".

killenb:
> "If and only if _activeException == true" is wrong, should be "If and only if _activeException != nullptr". It is not a boolean but a std::exception_ptr.

It explicitly converts to a bool. But maybe for the spec " != nullptr" is clearer.

killenb:
> 2. "Decorators must not swap a nullptr into the target because you could swap an active exception out of it." - What ensures that the outer layer does not get an exception swapped out when swapping in a real exception?

I don't understand this question. Only outer layers swap to the inner one (decorator to its target), not the other way around.

What can happen in a transfer group that the shared low level element still has the exception from the first high level element, which is then swapped into the second high level element (not a nullptr). I think this cannot lead to any problems because the innermost layer is throwing.

Maybe it is better not to swap, but to copy to the target and then set the own exception to nullptr. This is clean in all cases.

mhier:
> > 1. "Decorators must not swap a nullptr into the target because you could swap an active exception out of it." - What ensures that the outer layer does not get an exception swapped out when swapping in a real exception?
> 
> I don't understand this question. Only outer layers swap to the inner one (decorator to its target), not the other way around.

Swapping always works both ways. That's what you have referred to in the specs in this sentence?


> What can happen in a transfer group that the shared low level element still has the exception from the first high level element, which is then swapped into the second high level element (not a nullptr). I think this cannot lead to any problems because the innermost layer is throwing.

Ah right, in this case we even must not overwrite the exception with a nullptr, because it shall be rethrown. Right? This should be mentioned in the comment.


> Maybe it is better not to swap, but to copy to the target and then set the own exception to nullptr. This is clean in all cases.

I also though that, but still we must not overwrite an exception in the target with a nullptr. In any case, it is important that _activeException is nullptr after an exception has been given to the target, because the target might decide not to throw. In a TransferGroup we can have multiple exceptions at the same time, hence we cannot rely of having a nullptr in the target even if we have an exception.

We should provide a function to transport the exception into the target, because it is more than one line of code and has to be made identical for each decorator-like implementation.

killenb:
> We should provide a function to transport the exception into the target, because it is more than one line of code and has to be made identical for each decorator-like implementation.

I will introduce a function and make the _activeException protected again.

killenb:
I completely re-phrased C.2.3 after the introduction of TransferElement::setActiveException(). Please re-check.

doc/spec_TransferElement.dox
```diff
@@ -263,8 +263,8 @@ This documnent is currently still **INCOMPLETE**!
...
-  - \anchor transferElement_C_2_3 2.3 If and only if _activeException == true, decorators must swap the _activeException with its target before delegating to the target's postXxx(). This swaps the
-    active exception into the target, which is now responsible for handling it, and swaps in a nullptr from the target, so the decorator level does not throw any more \ref transferElement_comment_C_2_3a "(*)". \ref transferElement_comment_C_2_3b "(*)"
+  - \anchor transferElement_C_2_3 2.3 Decorators must delegate the TransferElement::_activeException to their target using TransferElement::setActiveException(). If the delegated exception is not nullptr,
+  the targets _active exception will be replaced by it, and the delegated exeption, which is handed by reference, is set to nullptr \ref transferElement_comment_C_2_3a "(*)". The target is now responsible for handling the exception. If the delegated exception already is nullptr, setActiveException() has no effect. A nullptr does not overwrite an active exception in the target \ref transferElement_comment_C_2_3b "(*)". \ref transferElement_comment_C_2_3c "(*)"
```
mhier, line 267: Shouldn't there be an exception that the exception is not delegated if it was thrown by the decorator itself?

mhier: (btw: It is helpful if we have in the source code one line per paragraph. There is an extra newline character here. This makes it harder to read if that newline does not accidentally match roughly the automatic line wrapping of the editor or the github display...)

killenb: Yes, that's 2.3.1, just below.

killenb: oh, it still says swap there. I'll fix that

killenb: I just fixed 2.3.1 and removed the newline. Did not notice it in my browser. Must have aligned perfectly with the line break.

doc/spec_TransferElement.dox
```diff
@@ -263,9 +263,8 @@ This documnent is currently still **INCOMPLETE**!
...
-    - 2.3.1 Decorators which are throwing themselves in doPreXxx(), before delegating to the target preXxx(), must remember this. If they did throw in doPreXxx(), then in doPostXxx() they do not swap the exception into the target and do not delegate to postXxx(). The exception is then re-thrown by the decorator's postXxx().
+  - \anchor transferElement_C_2_3 2.3 Decorators must delegate the TransferElement::_activeException to their target using TransferElement::setActiveException(). If the delegated exception is not nullptr, the targets _active exception will be replaced by it, and the delegated exeption, which is handed by reference, is set to nullptr \ref transferElement_comment_C_2_3a "(*)". The target is now responsible for handling the exception. If the delegated exception already is nullptr, setActiveException() has no effect. A nullptr does not overwrite an active exception in the target \ref transferElement_comment_C_2_3b "(*)". \ref transferElement_comment_C_2_3c "(*)"
```
mhier, line 266: Typo, please correct:  `_active exception` -> ` _activeException`

mhier: > and the delegated exeption, which is handed by reference, is set to nullptr`

I would change the wording:

> and the _activeException of the decorator is set to nullptr.

The "delegated exception" is in my opinion the exception itself, not the _activeException pointer to it. The exception is delegated to the target and hence not set to nullptr...

mhier: > If the delegated exception already is nullptr, setActiveException() has no effect. A nullptr does not overwrite an active exception in the target.

This is duplicate, which IMO makes it harder to understand (because one assumes the second sentence has a different meaning). Also, I would also here replace "delegated exception" with "_activeException".

mhier: And a general wording comment: I would prefer to avoid using "must" and "will" etc., because these words always imply a certain point of view. So I prefer:
> Decorators delegate the TransferElement::_activeException

and

> If the delegated exception is not nullptr, the targets _active exception is replaced by it

This makes the statements neutral and does not imply whether the reader is implementing a decorator, using one, or trying to understand what a decorator does to his backend implementation.

doc/spec_TransferElement.dox
```diff
@@ -263,9 +263,8 @@ This documnent is currently still **INCOMPLETE**!
...
+    - 2.3.1 Decorators which are throwing themselves in doPreXxx(), before delegating to the target preXxx(), must remember this. If they did throw in doPreXxx(), then in doPostXxx() they do not call setActiveException() of the target and do not delegate to postXxx(). The exception is then re-thrown by the decorator's postXxx().
```
mhier, line 267: `must remember this` -> `remembers this`

doc/spec_TransferElement.dox
killenb: I implemented all changes.
It would be nice to understand what is implemented in DeviceAccess already, and which part of the spec has to be followed when writing your own decorators, backends and transfer elements. But maybe the "must, will" etc. is not the right tool. I once worked on a document which started with an explanation "In this document should means this.., must means this, shall means something different". It was horrible to read and understand.

doc/spec_TransferElement.dox
killenb: done

doc/spec_TransferElement.dox
mhier: "A does something" implies in the same way that A is responsible for doing so as "A must do something". IMO the second wording is just very hard to read if one is currently not interested in the requirements to A. Usually there are at least two sides for each statement. E.g. "A gives something to B if condition" implies many things: "A must give something to B under the condition", "B must receive something from A under the condition", "B can expect A to give something under the condition" (which is kind of equal to "A will give something to B under the condition"), and usually also that all this does not happen if the condition is not met. This kind of gets lost, without really adding something valuable.

killenb: Right. We should use the simplest unambiguous wording. Adding extra words or fancy wording does not help.
In contrast to a letter or publication each sentence can, even should, always use the same structure. It does not have to be 'nice' to read, but as easy to understand as possible.

PR #160, opened 2020-06-17
Transfer element v1.0 rc7

killenb:
- B.9 Clarified that DeviceBackend::setException() is only used with runtime_errors
- B.10.2 Added info about TransferElement::setExceptionBackend() TransferElement::_exceptionBackend
- B.10.3 TransferElement::isFunctional() must return false while there is an active exception.


doc/spec_TransferElement.dox
```diff
@@ -239,7 +243,8 @@ This documnent is currently still **INCOMPLETE**!
 - 8.5.2 As the asynchonous mechanism and a synchronous read are two idependent channels there are potential race condition, depending on the exact protocol.
   The backend has to avoid this if possible. If it cannot be avoided, the implementation must make sure that the last value in the queue is the newest value, and this is not dopped or missed, even if the values before are not in order or send twice.
 
-- 9. It does not matter if the exception occured in an asynchronous or synchronous read, or in a write operation.
+- \anchor transferElement_comment_B_9a \ref transferElement_B_9 "9." The ChimeraTK::runtime_error is the only exception that conceptually is recoverable by calling DeviceBackend::open(). ChimeraTK::logic_error is a programming or configuration mistake. The boost::numeric::bad_numeric_cast conceptually also is a logical error (wrong data type) which only shows at run time when the data is overflowing. The boost::thread_interrupted is used when destroying the DeviceBackend object, so there is nothing to recover afterwards.
```
mhier, line 246: > The boost::thread_interrupted is used when destroying the DeviceBackend object, so there is nothing to recover afterwards.

I think this is not the right argument. The idea is to keep the backend and accessor functional even if thead_interrupted is thrown - otherwise we could greatly simplify the specs here (just throw it completely unhandled). One could use the thread_interrupted exception also in different contexts. It is simply the equivalent of boost::thread::interrupt(). There is nothing to recover, because nothing is broken. We merely interrupt the blocking function to continue with something else.

doc/spec_TransferElement.dox
```diff
@@ -248,6 +253,7 @@ This documnent is currently still **INCOMPLETE**!
 - 9.2.4 Avoid race conditions here. The call to Device::activateAsyncRead() usually is done from a different thread than the transfer which caused the exception.
 
 - \anchor transferElement_comment_B_10_1_3 \ref transferElement_B_10_1_3 "10.1.3" For instance a watchdog which monitors a reference register to detect firmware reboots that is not seen on the transport layer can trigger the exception state to inhibit asynchronous transfers while running a recovery procedure.
+- \anchor transferElement_comment_B_10_2_4 \ref transferElement_B_10_2_4 "10.2.4" This functionality is already implemented in the NDRegisterAccessorDecorator base class. It does not apply for decorator-like TransferElements which are created by backends, because the backend has to call TransferElement::setExceptionBackend afte the creation anyway (see \ref transferElement_B_10_2_1 "10.2.1").
```
mhier, line 256: afte -> after

doc/spec_TransferElement.dox
killenb: Can boost::thread_interrupted only be thrown as a result of calling TE::interrupt()?

mhier: yes

mhier: It was written at some point explicitly that it may only be thrown when interrupt() is called. This was removed when we moved it into the base class (was in section C before). Now we have 8.6 stating it is done like that, and C.1 does not list it as an allowed exception.

killenb: OK. I re-phrased the last sentense:
```
The boost::thread_interrupted is thrown when TransferElement::interrupt() is called. It means that a blocking read() has been interrupted on request, not that the device is malfunctioning and would need recovery.
```

killenb: I just pushed the changes

doc/spec_TransferElement.dox
killenb: done

killenb:
The conflit with the master is intentional. I increased the version number there to V1.0RC8WIP. It contains the "new, unreviewed sections" part which has to be checked before going to the final release. I will resolve the conflict when merging.

doc/spec_TransferElement.dox
```diff
@@ -158,8 +158,8 @@ This documnent is currently still **INCOMPLETE**!
...
     - \anchor transferElement_B_8_6_5 8.6.5 Calling TransferElement::interrupt() without AccessMode::wait_for_new_data will cause a ChimeraTK::logic_error. [\ref testTransferElement_B_8_6_5 "T"]
 
-- 9. If one transfer element of a device has seen an exception(*), all other transfer elements of the same device must also be aware of this.
-   - \anchor transferElement_B_9_1 9.1 The transfer element which detects the exception reports it to its TransferElement::exceptionBackend by calling DeviceBackend::setException \ref transferElement_comment_B_9_1 "(*)" (cf. \ref transferElement_B_10_1 "10.1")
+- 9. \anchor transferElement_B_9 If one transfer element of a device has seen a ChimeraTK::runtime_error \ref transferElement_comment_B_9a "(*)", all other transfer elements of the same device must also be aware of this. \ref  transferElement_comment_B_9b "(*)"
+   - \anchor transferElement_B_9_1 9.1 The transfer element which detects the ChimeraTK::runtime_error reports it to its TransferElement::_exceptionBackend by calling DeviceBackend::setException() \ref transferElement_comment_B_9_1 "(*)" (cf. \ref transferElement_B_10_1 "10.1")
```
phako, line 162: I'm kind of missing an explanation what the "_exceptionBackend" is. I assume it is the back-end that handles the exception (so in normal cases the device backend, for LMAP the LMAP backend itself)?

doc/spec_TransferElement.dox
```diff
@@ -239,7 +243,8 @@ This documnent is currently still **INCOMPLETE**!
 - 8.5.2 As the asynchonous mechanism and a synchronous read are two idependent channels there are potential race condition, depending on the exact protocol.
   The backend has to avoid this if possible. If it cannot be avoided, the implementation must make sure that the last value in the queue is the newest value, and this is not dopped or missed, even if the values before are not in order or send twice.
 
-- 9. It does not matter if the exception occured in an asynchronous or synchronous read, or in a write operation.
+- \anchor transferElement_comment_B_9a \ref transferElement_B_9 "9." The ChimeraTK::runtime_error is the only exception that conceptually is recoverable by calling DeviceBackend::open(). ChimeraTK::logic_error is a programming or configuration mistake. The boost::numeric::bad_numeric_cast conceptually also is a logical error (wrong data type) which only shows at run time when the data is overflowing. The boost::thread_interrupted is thrown when TransferElement::interrupt() is called. It means that a blocking read() has been interrupted on request, not that the device is malfunctioning and would need recovery.
```
phako, line 246: That to me sounds that only runtime_error is recoverable form only by open(), the others by other means not further specified.
```
ChimeraTK::runtime_error is the only exception that conceptually is recoverable. This is done by calling DeviceBackend::open(). Other exceptions we expect are:
 - ctk::logic_error which is a programming or configuration error and cannot be recovered
 - boost::numeric_bad_numeric_cast which is conceptually identical to a logic error (wrong data type). It only shows at runtime when the data is overflowing
 - boost::thread_interrupted is used to cleanly shut down threads. It is not an error, but a way to interrupt a blocking read() on request, by calling TransferElement::interrupt()
```
The explanation of boost::thread_interrupted might need a bit of more refinement

doc/spec_TransferElement.dox
killenb: As 9.1 say it is the backend to which you report the exception. And yes, it is responsible for executing the actions 9.2, 9.3, 9.4.
This is different from the backend which created the TransferElement. For instance if there are two PCIe-backends in one LNM backend. The TE is directly generated by the PCIe and given out by the LNM.
Now if one of the two PCIes sees an exception, the application only known it talking to the LNM, the only one device it known. So all accessors must be in error state. Hence the exception is reported to the LNM (the exceptionBackend) which delegates the call to all its target backends. Although the Accessor has the PCIe als its internal target on which it calls NumericAddressBackend::read() etc.

doc/spec_TransferElement.dox
killenb: > boost::thread_interrupted is used to cleanly shut down threads.
That's what it can be used for. But it can also be used do just cancel an read request. For instance you wait for a variable while a calibration is running. If the calibration succeeds without any thing being read, you don't need the value any more. You call interrupt() on that read request and continue with something else.
I would not put what we use it for into the spec.

killenb: > It is not an error, but a way to interrupt a blocking read().
No, the exception is not the way to interrupt the read. The call to TE::interrupt is.
I also would not use the tern "error" in this case. Errors in the sense of logical errors / programming mistakes are conceptually not recoverable. The point here that it is not a "fault", i.e. malfunctioning which could be recovered.

doc/spec_TransferElement.dox
phako: So mayebe just put he "may be different from the back-end that created the TE, e.g. in aggregating back-ends such as LNM! in a comment?

doc/spec_TransferElement.dox
phako: Fine with me. The main point was that it sounded as if those were faults recoverable by other means than open()

doc/spec_TransferElement.dox
killenb: This will be in the changes that Martin requested anyway. I would postpone it to RC8

doc/spec_TransferElement.dox
killenb: Ah. that sentence I can put in. And maybe do the bullet points for better overview.

PR #164, opened 2020-07-16
Wip/mhier/transfer element spec rc10

mhier:


doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+      - \anchor transferElement_B_12_10_1_2 12.10.1.2 Any pre-condition which is not met will immediately result in a thrown ChimeraTK::logic_error
+      - 12.10.1.3 The exceptions from \ref transferElement_B_12_10_1_1 "12.10.1.1" and \ref transferElement_B_12_10_1_2 "12.10.1.2" are not prioritised by type. The first detected error throws.
+    - 12.10.2 ChimeraTK::runtime_error exceptions thrown by the low-level elements are propagated to the corresponding high-level elements so they are seen by their postXxx() functions (c.f. \ref transferElement_B_6_3 "6.3")
+    - 12.10.3 postXxx() is called for all high level elements, even if some postXxx() throw exceptions (c.f. \ref transferElement_B_5 "5").
+      - 12.10.3.1 If exceptions have been caught from any postXxx(), an exception is thrown after all of them have been executed.
+      - 12.10.3.2 If more than one exception type is detected, the exception type with the highest priority is thrown:
+        - ChimeraTK::logic_error is thrown before any transfer is actually started, so it implicitly has highest priority.
```
killenb, line 273: This section is about generating a single exception out of the many exceptions that can be (re-)thrown in postXxx. So there are no logic errors here. Like boost::thread_interrupted, they cannot occur here.

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+        - If a ChimeraTK::runtime_error is thrown by a transfer, it will be passed on to the application.
+        - boost::numeric_cast is lowest priority, since in case of a ChimeraTK::runtime_error the content of the application data buffer is unchanged (see \ref transferElement_B_12_11 "12.11").
```
killenb, line 275: > boost::numeric_cast is lowest priority

This is wrong. boost::numeric_cast is not recoverable. It conceptually is like a logic error, except that is shows up too late. It must not be hidden by a runtime_error. It must have higher priority.

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+      - \anchor transferElement_B_12_10_1_2 12.10.1.2 Any pre-condition which is not met will immediately result in a thrown ChimeraTK::logic_error
+      - 12.10.1.3 The exceptions from \ref transferElement_B_12_10_1_1 "12.10.1.1" and \ref transferElement_B_12_10_1_2 "12.10.1.2" are not prioritised by type. The first detected error throws.
+    - 12.10.2 ChimeraTK::runtime_error exceptions thrown by the low-level elements are propagated to the corresponding high-level elements so they are seen by their postXxx() functions (c.f. \ref transferElement_B_6_3 "6.3")
+    - 12.10.3 postXxx() is called for all high level elements, even if some postXxx() throw exceptions (c.f. \ref transferElement_B_5 "5").
+      - 12.10.3.1 If exceptions have been caught from any postXxx(), an exception is thrown after all of them have been executed.
+      - 12.10.3.2 If more than one exception type is detected, the exception type with the highest priority is thrown:
+        - ChimeraTK::logic_error is thrown before any transfer is actually started, so it implicitly has highest priority.
+        - If a ChimeraTK::runtime_error is thrown by a transfer, it will be passed on to the application.
```
killenb, line 274: This is confusing. And there can be multiple. I will propose something below which hopefully is clearer.

doc/spec_TransferElement.dox
killenb: Ah, I just realised there is no conflict. They cannot appear at the same time. See comment to 12.11.1

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+        - boost::numeric_cast is lowest priority, since in case of a ChimeraTK::runtime_error the content of the application data buffer is unchanged (see \ref transferElement_B_12_11 "12.11").
+        - boost::thread_interrupted cannot take place inside TransferGroups, since they only occur with AccessMode::wait_for_new_data.
+  - \anchor transferElement_B_12_11 12.11 If a ChimeraTK::runtime_error is thrown, the content of the application buffers of all elements is not changed. (c.f. \ref transferElement_B_6_4 "6.4")
+    - 12.11.1 The TransferGroup calls postRead() of all elements with `updateDataBuffer = false`, if any element has thrown a ChimeraTK::runtime_error.
```
killenb, line 278: 12.11.1.1 This means if there has been a runtime_error there cannot be a boost::bad_numeric_cast. On the other hand, if there is a boost::bad_numeric_cast, there cannot be a runtime_error because postXxx() only allowed to re-throw them, not raise them.

killenb: see C.4.2

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+      - \anchor transferElement_B_12_10_1_1 12.10.1.1 If any ChimeraTK::runtime_error occurs while checking the pre-conditions, it is immediately passed on to the application.
+      - \anchor transferElement_B_12_10_1_2 12.10.1.2 Any pre-condition which is not met will immediately result in a thrown ChimeraTK::logic_error
+      - 12.10.1.3 The exceptions from \ref transferElement_B_12_10_1_1 "12.10.1.1" and \ref transferElement_B_12_10_1_2 "12.10.1.2" are not prioritised by type. The first detected error throws.
+    - 12.10.2 ChimeraTK::runtime_error exceptions thrown by the low-level elements are propagated to the corresponding high-level elements so they are seen by their postXxx() functions (c.f. \ref transferElement_B_6_3 "6.3")
+    - 12.10.3 postXxx() is called for all high level elements, even if some postXxx() throw exceptions (c.f. \ref transferElement_B_5 "5").
+      - 12.10.3.1 If exceptions have been caught from any postXxx(), an exception is thrown after all of them have been executed.
```
killenb, line 271: Proposal for 12.10.3.1
```
* 12.10.3.1 All exception messages of the postXxx() calls are collected.
```

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+      - 12.10.3.2 If more than one exception type is detected, the exception type with the highest priority is thrown:
```
killenb, line 272: Proposal:
```
* 12.10.3.2 As only either runtime_errors or boost::bad_numeric_cast can occur (see 12.11.1.1), it is clear which exception type to throw (see 12.11.1.1).
```

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+  - 12.4 A TransferGroup is becomes read-only (cf. TransferGroup::isReadOnly()), if at least one of the member TransferElements is read-only.
+  - 12.5 Adding the same TransferElement twice is not allowed (logic_error).
+  - 12.6 Adding one TransferElement to two different TransferGroups is not allowed (logic_error).
+  - 12.7 Adding TransferElements with AccessMode::wait_for_new_data is not allowed (logic_error).
+  - 12.8 Adding a copy of the a TransferElement which is already in the group (returned by two different calls to Device::getZzzRegisterAccessor() resp. DeviceBackend::getRegisterAccessor() with the same arguments) will replace one of the identical TransferElement with a decorator which copies the data from the other.
+    - 12.8.1 This avoids an unnecessary transfer of identical data.
+    - 12.8.2 This is only possible, if the TransferElement is added by passing a TransferElementAbstractor to TransferGroup::addAccessor().
```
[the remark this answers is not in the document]
mhier: You mean why it is only possible when an abstractor is used? Because you need some kind of "holder" so you can replace the highest level. This is a logical consequence of the interface. The behaviour is influenced by it so it needs to be mentioned.
"Implementation detail" means it can be done differently without anyone noticing. This is not the case here.

doc/spec_TransferElement.dox
mhier: This is what it says, isn't it?

doc/spec_TransferElement.dox
mhier: Only one exception can be thrown at a time in C++. I want to avoid the need for a sophisticated to prioritise exception types at that point, because there is no benefit from that. You can always have first a runtime_error, because no communication is possible, and when this is resolved a logic_error, because only then the information to decide about the logic_error is available. Keep in mind, the TransferGroup does not reveal which accessor has thrown.

doc/spec_TransferElement.dox
mhier: This is a comment at most. IMO there is already too much about stupid numeric_cast errors in the spec. We want to get rid of them.

doc/spec_TransferElement.dox
mhier: Better:
- 12.10.3.1.1 The thrown exception will contain all messages from the caught exceptions.

mhier: Actually, I think this is a bad idea. We should merely keep the first message. Otherwise we will get in the vast majority of cases very very long error messages which are repeating the same error many times. Probably it will be completely unreadable and also create displaying issues due to too long strings. The chance that there is a real value in the additional information which can actually be found is rather slim.

TransferGroups may have hundreds of variables of usually very few devices. Most of the time all (single device) or a good fraction (few devices) of the variables will have the same error simultaneously. Even if two devices fail at the same time, who is supposed to find the actually different message in that pile of messages?

We could invent some sophisticated mechanism to filter duplicates, but this may be spoiled by the actually good practise of including the variable name in the message (-> we should probably even require this from all backends, helps debugging a lot). It's not worth doing this. Just keep the first message and that's it.

doc/spec_TransferElement.dox
killenb: Well, the logic error has no priority (neither high nor low) in this context because it cannot appear. But the whole thing goes away with the requested change to 12.10.3.2. There are no priorities any more.

doc/spec_TransferElement.dox
mhier: Hence:

> * 12.10.3.1.1 The thrown exception will contain the message of the first caught exception.

doc/spec_TransferElement.dox
mhier: Maybe remove that entire block. Ignore stupid cast errors. Then only runtime_errors are left. Put a comment that thread_interrupted cannot happen. That's it. Everything else is unnecessary bloat.

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,50 @@ This documnent is currently still **INCOMPLETE**!
...
+  - 12.2 Adding TransferElements which cannot merge their transfers to the same TransferGroup is always allowed.
+  - 12.3 If TransferElements have been added to a TransferGroup, individual read/write operations of the TransferElements are no longer allowed and will throw a ChimeraTK::logic_error. Read/write operations are then only possible through the TransferGroup.
+  - 12.4 A TransferGroup is becomes read-only (cf. TransferGroup::isReadOnly()), if at least one of the member TransferElements is read-only.
+  - 12.5 Adding the same TransferElement twice is not allowed (logic_error).
+  - 12.6 Adding one TransferElement to two different TransferGroups is not allowed (logic_error).
+  - 12.7 Adding TransferElements with AccessMode::wait_for_new_data is not allowed (logic_error).
+  - 12.8 Adding a copy of the a TransferElement which is already in the group (returned by two different calls to Device::getZzzRegisterAccessor() resp. DeviceBackend::getRegisterAccessor() with the same arguments) will replace one of the identical TransferElement with a decorator which copies the data from the other.
```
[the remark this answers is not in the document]
mhier: Actually, I am not sure if we really use this. Today I wouldn't implement this any more, because it's probably not worth it. Now we have it in, and it may happen e.g. if one has two logical name mapping registers redirecting to the same target.

doc/spec_TransferElement.dox
killenb: Ok, we can just let them through and terminate the application. If that's acceptable until they have been removed this is also fine.

doc/spec_TransferElement.dox
killenb: You are right. But it is not the first caught exception. I'll explain why:

* What actually will happen with the new requirement of setException is that after the first transfer gave a useful message, all others will say "follow-up error", which we don't want.
* So we have to get the message from the first transfer which has seen an exception. But it is not thrown, it happens under the hood. We only see that _activeException is not nullptr.
* We cannot get the first thrown exception from the post reads. They might happen in a different order than the transfers.

doc/spec_TransferElement.dox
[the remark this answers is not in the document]
killenb: Ah, the highest element is exchanged. Then it is clear why this can only be done when you use the abstractor.

But why is this treated differently to another TE without abstractor with the same target being added, where they can share the same low level TE?

Ah, I see, this also works if they cannot share the transfer because they don't have low level TEs and do not support replacing each other in the low level. This should go to a comment.

Still I don't see why you would want to add different TEs to the same data multiple times.

doc/spec_TransferElement.dox
killenb: > because only then the information to decide about the logic_error is available

I don't think the information about logic errors can be done after a transfer has thrown a runtime_error. We are not talking about the runtime_errors in the checking pre-conditions phase. Those have been dealt with at this point. It's only about exceptions from the transfer phase.

killenb: Ah, now I know what you mean. In the next call to read. But that's a new round.

doc/spec_TransferElement.dox
killenb: Yes, that's even simpler. But with your proposal to only throw the first exception it will look like this: (not spec, what happens)

* execute transfer phase. Exceptions are delayed.
* call all postXxx() with their correct exceptions (and updateDataBuffer = false for all if any exception in transfer phase)
* catch all runtime_errors of the postXxx() phase and ignore them (logic and tread_interrupted cannot occur, numeric we don't care)
* Iterate all lowLevel TEs again and re-throw the first exception that is not nullptr.

doc/spec_TransferElement.dox
mhier: Yes it is the first caught/thrown exception. The original throwing happens in the transfer. Everything else is just rethrowing. How the TransferGroup organises itself to make sure that actually the originally first error is rethrown is an implementation detail.

So maybe it is clearer like this:

> 12.10.3.1.1 The thrown exception will contain the message of the first thrown exception.

doc/spec_TransferElement.dox
mhier: Ok I think I don't understand what your comment refers to here. Is this all maybe already discussed below in the other points? I am lost.

doc/spec_TransferElement.dox
mhier: Let's define that this is the intended behaviour at this point.

doc/spec_TransferElement.dox
mhier: Again I am confused about the many discussions about the same (?) topic here. Is anything in this thread what has not been mentioned/clarified by the one about 12.10.3.1?

doc/spec_TransferElement.dox
mhier: Let's converge this discussion in the thread about 12.10.3.1.

doc/spec_TransferElement.dox
mhier: Let's converge this discussion in the thread about 12.10.3.1.

doc/spec_TransferElement.dox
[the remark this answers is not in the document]
mhier: > Still I don't see why you would want to add different TEs to the same data multiple times.

It can happen even somewhat indirectly if two logical name mapping registers redirect to the same register. It's not a super important feature and I probably would not implement it any more today, but it has some usefulness and does not overcomplicate things too much, so I would keep it in.

It is actually not treated differently, there is something missing in the spec: A copy decorator is also used when replacing a transfer element on the inner levels 1:1, unless the implementation can guarantee that all peers accessing the same low-level element are keeping its application data buffer intact (i.e. they do not swap out the buffer). For that reason, NDRegisterAccessorDecorator::replaceTransferElement() creates a copy decorator, because it swaps the data even itself. I just found on the other hand that the LMNBackendBitAccessor::replaceTransferElement() does not create a copy decorator, which is a bug. So far it goes unnoticed, because usually the peers are only other LMNBackendBitAccessors, which do not swap out the data. It would go wrong, if the target accessor of a bit accessor would be used as a target of a decorator which swaps out the data, and by chance the decorators postRead is called first. Then the bit accessors would see wrong data. (Might still go unnoticed btw., because data would simply be outdated... Hard to catch!)

I will add a comment.

doc/spec_TransferElement.dox
killenb: Agreed.

doc/spec_TransferElement.dox
killenb: I don't know if what I am writing is helpful, because it it not only about this point. Here is my summary of all the discussions, without distinction to which point it belongs:

* boost::bad_numeric_cast is not treated
* boost::thread_interrupted cannot occur
* preConditons check that no ChimeraTK::logic_error can occur in preXxx, as described
* => only runtime_errors are left, can only primarily happen on transfer phase
* exceptions in the transfer phase are delayed,  so all postXxx are called with the correct exception
* If an exception is seen in the transfer phase, all postRead are called with UpdateDataBuffer = false
* All re-thrown exceptions from the postXxx-phase are discarded.
* The first detected exception from the transfer phase is re-thrown after all postXxx have been executed.

mhier: numeric_cast should be treated, so that all postReads are called. We just don't have to care about priorities because stuff will natively come in a well-defined order. Just throwing the first error (after all postXxx have been executed) is enough.

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,49 @@ This documnent is currently still **INCOMPLETE**!
...
+      - \anchor transferElement_B_12_10_1_1 12.10.1.1 If any ChimeraTK::runtime_error occurs while checking the pre-conditions, it is immediately passed on to the application.
+      - \anchor transferElement_B_12_10_1_2 12.10.1.2 Any pre-condition which is not met will immediately result in a thrown ChimeraTK::logic_error
+      - 12.10.1.3 The exceptions from \ref transferElement_B_12_10_1_1 "12.10.1.1" and \ref transferElement_B_12_10_1_2 "12.10.1.2" are not prioritised by type. The first detected error throws.
+    - 12.10.2 ChimeraTK::runtime_error exceptions thrown by the low-level elements are propagated to the corresponding high-level elements so they are seen by their postXxx() functions (c.f. \ref transferElement_B_6_3 "6.3")
+    - 12.10.3 postXxx() is called for all high level elements, even if some postXxx() throw exceptions (c.f. \ref transferElement_B_5 "5").
+      - 12.10.3.1 If exceptions have been caught from any postXxx(), an exception is thrown after all of them have been executed.
+      - 12.10.3.2 The thrown exception will be identical to the first thrown exception.
```
killenb, line 274: Sounds confusion. What is "the first thrown exception"?
Proposal:
The first exception that was seen by the transfer group in 12.9.2 or 12.9.3 is re-thrown (*)

(*) "Was seen" means either as _activeException of an low level TE after calling the transfer step, or a bad_numeric_cast caught in postRead().

mhier: I can do that, but I think you are merely confusing two implementations which have the same behaviour. You can either check which exception has appeared first in the transfer phase in _activeException of the low-level elements, or you can build some smart try-catch block around the postXxx() calls and rethrow the first thrown exception. Your suggestion implies the first implementation, while just stating "first thrown exception" is neutral to this.

If I am wrong about this: what else do you think this can be confused with?

mhier: (I actually find your suggestion more confusing... It's overcomplicating a simple idea and implies a specific implementation.)

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,49 @@ This documnent is currently still **INCOMPLETE**!
...
+      - 12.1.1.1 If a merge happens and consistency of the data can be guaranteed in a read operation, the VersionNumber of the corresponding TransferElements is set to the same value.
+    - 12.1.2 Structure of TransferElements in TransferGroups:
+      - \anchor transferElement_B_12_1_2_1 12.1.2.1 A TransferElement, which is directly added to the TransferGroup, is called <i>high-level element</i>.
+      - \anchor transferElement_B_12_1_2_2 12.1.2.2 Decorator-like implementations use one or more <i>low-level elements</i> which perform the actual register access. All low-level elements of one TransferElement are returned by TransferElement::getHardwareAccessingElements().
+      - 12.1.2.3 Low-level elements are not decorator-like, since they directly access the device.
+      - 12.1.2.4 There can be any number of decorator-like intermediate elements. The sum of all elements on all levels of one high-level element including the low-level elements but excluding the high-level element itself is called <i>internal elements</i>. All internal elements of one TransferElement are returned by TransferElement::getInternalElements().
```
phako, line 243: More of a comment: This might also make sense to put it into some kind of TransferGroup glossary or detailed description, outside of this spec. This is important terminology imho and should not be buried inside the TE spec

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,49 @@ This documnent is currently still **INCOMPLETE**!
...
+      - 12.1.4.2 The given replacement can be extended such that the data of an internal element is included in the transfer; then the internal element can be replaced by the now-extended replacement.
+      - 12.1.4.3 If such replacement takes place, the behaviour visible to the application of the TransferElement does not change (exception: VersionNumbers in read opertions can reveal now-consistent data, see 12.1.1).
+      - 12.1.4.4 If no such optimisation is possible, the function simply does nothing.
+    - \anchor transferElement_B_12_1_5 12.1.5 TransferElement::mayReplaceOther() returns true, if the given TransferElement behaves identical to the one it is called on in every aspect, i.e. it accesses the same data on the same device, has the same UserType and the same AccessModeFlags, and (if applicable) uses the same data transformations (like fixed point conversion). It returns false otherwise. This is used by implementations of \ref transferElement_B_12_1_4_1 "12.1.4.1".
```
phako, line 252: That B.12.1.5 might be more suited in the TransferElement description itself and then referenced here.

doc/spec_TransferElement.dox
killenb: Well, the comment might be confusing. And actually, to implement it correctly you have to remember the first _activeExeption in the transfer phase (only those have the correct order) AND you need the try catch block for the bad_numeric_cast.

But that's not what I don't like. The wording is not clear.
```
The thrown exception will be identical to the first thrown exception.
```
I don't like the first *thrown* exception, especially after "the thrown exception". It's the first caught exception when I am inside the group code. And the "will be identical to" also does not help to make it clearer. In fact, it is the same exception, thrown with std::rethrow_exception.

How about this:
```
The first exception that has been caught by the transfer group in 12.9.2 or 12.9.3 is re-thrown.
```
without the comment. We could in fact call readTransfer() directly and catch in the group, and not call handleTransferException and look at _activeException. But that's implementation detail, it does not belong into the spec.

mhier: Ok I will change it to:

> The first exception that has been caught in 12.9.2 or 12.9.3 is re-thrown.

Then it's neutral w.r.t. the implementation.

mhier: 12.9.2 is for runtime_error and 12.9.3 is only for numeric_cast error, right?

doc/spec_TransferElement.dox
```diff
@@ -213,9 +232,49 @@ This documnent is currently still **INCOMPLETE**!
...
+      - \anchor transferElement_B_12_10_1_1 12.10.1.1 If any ChimeraTK::runtime_error occurs while checking the pre-conditions, it is immediately passed on to the application.
+      - \anchor transferElement_B_12_10_1_2 12.10.1.2 Any pre-condition which is not met will immediately result in a thrown ChimeraTK::logic_error
+      - 12.10.1.3 The exceptions from \ref transferElement_B_12_10_1_1 "12.10.1.1" and \ref transferElement_B_12_10_1_2 "12.10.1.2" are not prioritised by type. The first detected error throws.
+    - 12.10.2 ChimeraTK::runtime_error exceptions thrown by the low-level elements are propagated to the corresponding high-level elements so they are seen by their postXxx() functions (c.f. \ref transferElement_B_6_3 "6.3")
+    - 12.10.3 postXxx() is called for all high level elements, even if some postXxx() throw exceptions (c.f. \ref transferElement_B_5 "5").
+      - 12.10.3.1 If exceptions have been caught from any postXxx(), an exception is thrown after all of them have been executed.
```
killenb, line 273: With the new wording of 12.10.3.2 this is obsolete. postXxx() is 12.9.3.

Should we remove 12.10.3.1 and say transfer phase and postXxx instead of 12.9.2 and 12.9.3?

PR #178, opened 2020-09-16
Pr/mhier/spec transferelement recovery versionnumber

mhier:
See ChimeraTK/DeviceAccess-DoocsBackend#45. This requirement might apply to other backends as well, so it should go to the general spec.

Important: When this PR is accepted, please create a ticket to extend the UnifiedBackendTest for this requirement!


phako (approved):

mhier:
Actually, we need to rethink this.

I am still seeing VersionNumbers going backwards in case of exceptions+recovery. The new requirement from this PR ensures that the first version number returned by a read transfer is bigger than the exceptionVersion of the DeviceModule, but there still is no guarantee that the exceptionVersion is bigger that the version number of the last read transfer before the exception. This currently results in a logic_error in the ExceptionHandlingDecorator when it tries to return the exceptionVersion.

I am not sure what the best solution is. I will attempt some temporary work around to fix the logic_errors, but the behaviour will be probably somewhat inconsistent.

mhier:
And even the assumption that the requirement from this PR ensures that the first version after recovery is bigger than the exceptionVersion is wrong. I have put in an assertion into the ExceptionHandlingDecorator, and it fails.

I think the problem is that we can have multiple LogicalNameMappingBackends (and hence multiple DeviceModules) using the same backend. 

Probably we should not assume any particular relationship between the exceptionVersion and any of the version numbers before the exception and after the recovery. We need to deal with this situation differently somehow.

PR #192, opened 2020-11-06
Transfer element v1.0 rc13

killenb:
The boost::bad_numeric_cast exception has been removed. Instead the data is clamped into the data range than the data validity is set to faulty.
In addition, the ChimeraTK::logic_error is not allowed in the constructor any more. The data is always clamped.

With these changes the read and write direction have symmetric behaviour.

### DIFF
--- a/doc/spec_TransferElement.dox
+++ b/doc/spec_TransferElement.dox
@@ -1,7 +1,7 @@
 // put the namespace around the doxygen block so we don't have to give it all the time in the code to get links
 namespace ChimeraTK {
 /**
-\page spec_TransferElement Technical specification: TransferElement V1.0RC12
+\page spec_TransferElement Technical specification: TransferElement V1.0RC13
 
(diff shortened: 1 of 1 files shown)


doc/spec_TransferElement.dox
```diff
@@ -293,6 +294,8 @@ This documnent is currently still **INCOMPLETE**!
 
 
 ### (*) Comments ###
+- \anchor transferElement_comment_B_2_5_1  \ref transferElement_B_2_5_1 "2.5.1" Backends which don't support data validity flags (like the PCIe backend) just get the clampled value. The DOOCS backend however, which supports data validity, sets the outgoing data to invalid in addition.
```
mhier, line 297: I am actually not sure whether DOOCS supports flagging data as invalid in the writing direction... Maybe I would not name specific protocols here but keep the statement general.

mhier (approved):
One minor change requested in a comment.

phako (approved):

PR #206, opened 2021-04-14
Move bar index check to NumericAddressedBackend

patrislav1:
`NumericAddressedLowLevelTransferElement` shouldn't know details about which BAR values are valid and which ones are not. This patch delegates that check to the `NumericAddressedBackend` implementation. By default, it performs the same check that was done in `NumericAddressedLowLevelTransferElement` constructor before. Now a subclass of `NumericAddressedBackend` can override that check and re-define which BAR values are valid.

### DIFF
--- a/device_backends/include/NumericAddressedBackend.h
+++ b/device_backends/include/NumericAddressedBackend.h
@@ -30,6 +30,8 @@ namespace ChimeraTK {
 
     virtual void write(uint8_t bar, uint32_t address, int32_t const* data, size_t sizeInBytes) = 0;
 
+    virtual bool barIndexValid(uint32_t bar);
+
     virtual std::string readDeviceInfo() = 0;
 
     /**
--- a/device_backends/include/NumericAddressedLowLevelTransferElement.h
+++ b/device_backends/include/NumericAddressedLowLevelTransferElement.h
@@ -37,7 +37,7 @@ namespace ChimeraTK {
         boost::shared_ptr<NumericAddressedBackend> dev, size_t bar, size_t startAddress, size_t numberOfBytes)
     : TransferElement("", {AccessMode::raw}), _dev(dev), _bar(bar), isShared(false),
       _unalignedAccess(_dev->_unalignedAccess, std::defer_lock) {
-      if(bar > 5 && bar != 13) {
+      if(!dev->barIndexValid(bar)) {
         std::stringstream errorMessage;
         errorMessage << "Invalid bar number: " << bar << std::endl;
         throw ChimeraTK::logic_error(errorMessage.str());
--- a/device_backends/src/NumericAddressedBackend.cc
+++ b/device_backends/src/NumericAddressedBackend.cc
@@ -81,6 +81,11 @@ namespace ChimeraTK {
     write(retRegBar, retRegOff, data, retDataSize);
   }
 
+  // Default range of valid BARs
+  bool NumericAddressedBackend::barIndexValid(uint32_t bar) {
+    return bar <= 5 || bar == 13; 
+  }
+
   /********************************************************************************************************************/
 
   boost::shared_ptr<const RegisterInfoMap> NumericAddressedBackend::getRegisterMap() const { return _registerMap; }


killenb (changes requested):
In fact the implementation (<= 5 || 13) is only valid for pcie. It already is wrong for ReboT, which only allows bar 0.
I would make `barIndexValid()` pure virtual and force an implementation in the backends. Arrrgh, this breaks API backwards compatibility again.

I also don't like the "bar index" in the name. According to the current discussion about future extensions this might become a string. Ok, then the API changes again. But how about `isBarValid()`? (just a proposal. Like it better but still does not 100 % convince me)

Moving the validation of the bar to the backend is 100 % the right thing to do.

patrislav1:
Yes, the default implementation is a compromise to not break existing code.

According to the current discussion, the `bar` name itself is already questionable. (Don't know if there are already better candidates for a name).

Most of the API treats this value as `uint8_t`, so the type will probably also have to change.

patrislav1:
Would it make sense to choose a generic name such as `backend_arg` or `backend_opt`?
It's an argument that's passed transparently to the backend, and only the backend knows its meaning.
For `PcieBackend` it's an index, for `XdmaBackend` it's an enum, and others like `RebotBackend` silently ignore it.

And changing the type to string (or introducing an additional string-type option) could be beneficial for a future AXI backend. In the ARM AXI world, the backend will have to deal with named device files such as `hier_daq_arm_axi_bram_ctrl_0`, and a string value passed from the mapfile would allow it to easily address these device files.

mhier:
> Would it make sense to choose a generic name such as `backend_arg` or `backend_opt`?
> It's an argument that's passed transparently to the backend, and only the backend knows its meaning.
> For `PcieBackend` it's an index, for `XdmaBackend` it's an enum, and others like `RebotBackend` silently ignore it.

Isn't the data type fixed? All NumericAddressedBackends use the same map file parser and the same struct to store the data in the catalogue - so technically the XdmaBackend may cast it into an enum, but up to that point it still is a plain integer.

Also, it is not completely up to the backend to interpret this number. It is part of the address. Features like merging of transfers for adjacent addresses inside a TransferGroup check on this number and require the value to be identical to allow the transfers to merge. This may be only one example. Conceptionally the address in a NumericAddressedBackend consists of two integer values, and this needs to be kept in mind all the time.


> And changing the type to string (or introducing an additional string-type option) could be beneficial for a future AXI backend. In the ARM AXI world, the backend will have to deal with named device files such as `hier_daq_arm_axi_bram_ctrl_0`, and a string value passed from the mapfile would allow it to easily address these device files.

The device identification is not part of the map file. This device name will go into the CDD (ChimeraTK device descriptor) in the same way as the /dev node name currently is in there. These things should not be confused, otherwise many good concepts break...

Still, if needed we can of course extend the data type to a string, but this needs to be done carefully. As far as I remember, so far only equality is checked for the bar number outside the actual backend implementation, so this should work also for strings. Still, string comparison is more expensive, so we need to make sure it s not done in time critical places (the creation of a TransferGroup is not considered time critical...).

patrislav1:
> Isn't the data type fixed? All NumericAddressedBackends use the same map file parser and the same struct to store the data in the catalogue - so technically the XdmaBackend may cast it into an enum, but up to that point it still is a plain integer.

Yes, at the moment it's passed as a `uint8_t`. However it's interpreted differently (or sometimes not at all) depending on the backend.

> The device identification is not part of the map file. This device name will go into the CDD (ChimeraTK device descriptor) in the same way as the /dev node name currently is in there.

Ok, then a AXI backend will probably need a separate map file for each FPGA IP.

mhier:
> Yes, at the moment it's passed as a `uint8_t`. However it's interpreted differently (or sometimes not at all) depending on the backend.

In which way is this different to the byte address? The NumericAddressedBackend defines an address consisting of two parts, and which values are valid and what they exactly mean is of course depending on the backend, and btw. also depending on the device.

My point is: the backend is not completely free to define the meaning of the bar number, since it is part of the address and defines the address space to which the byte address is relative. Hence names like `backend_arg` or `backend_opt` are too generic. "BAR" is not so bad after all, but the term is somewhat coupled to PCIe and also has "register" in its name, which is untrue e.g. for Modbus. Maybe `address_space_index` would describe it better (as long as we stick to an integer - which I would prefer as long as it is possible).

killenb:
> Isn't the data type fixed? All NumericAddressedBackends use the same map file parser

Yes. In the moment the map file parser returns an int8_t, and I would not change it at the moment. 

Everything that I write below is part of a discussion for a future extension. Nothing to change now because those are only early ideas:

> Also, it is not completely up to the backend to interpret this number. It is part of the address.

This is part of the current discussion. To be exact the numeric address is in a virtual address space which is described by something like a "magic number" (for instance pcie 0-5 means PCI BAR 0-5, 13 means DMA etc.) The idea is to change this to a more generic identifier for the address space which can transport information to the backend. Example: Some registers can in future be read via memory map or DMA transfer. If you have a single word, you don't want to go through the effort to do a DMA transfer with interrupt etc., but if you have the same register in a transfer group with many others, you might want a DMA transfer because reading via memory map is less efficient. This can only be decided by the backend.

The alternative that is being discussed is to keep magic numbers and have meta information in the map files what those magic numbers mean.

In addition the firmware group wants the map file to work with different transport mechanism (PCIe if you are on a CPU and AXI if you have SytemOnChip). Here some meta information like additional offsets in the AXI address space have to be transported to the backend. This is the type of information that is has to go to the map file, because it depends on the bit-file.

As @mhier already wrote, the device node where it shows on the host computer goes to the dmap file and is not part of the map file.

killenb:
To move forward with this merge request I see two points that have to be answered:

1. Is the name `barIndexValid()` ok or do we want to change it?
2. Do we have a default implementation to keep compatibitily or do we have a pure virtual function? The latter is the cleaner solution but breaks the API. 
2.a If we have a default implementation, which one is it? I would vote for only allowing BAR 0. If the backends wants to use magic numbers it has to override.

Maybe a side question: Is this check needed at that place? Or is a virtual function all at that place already too expensive?

mhier:
> This is part of the current discussion. To be exact the numeric address is in a virtual address space which is described by something like a "magic number" (for instance pcie 0-5 means PCI BAR 0-5, 13 means DMA etc.) The idea is to change this to a more generic identifier for the address space which can transport information to the backend. Example: Some registers can in future be read via memory map or DMA transfer. If you have a single word, you don't want to go through the effort to do a DMA transfer with interrupt etc., but if you have the same register in a transfer group with many others, you might want a DMA transfer because reading via memory map is less efficient. This can only be decided by the backend.

Again: The bar number must not be abused to encode something which is logically not part of the address. Of course, a backend could define address spaces which have some special meaning, like the 0xD address space for DMA. But the moment these address spaces have an overlap, the NumericAddressedBackend may be confused because it does not know this. So the scenario you describe does not work that way, or at least would be limited to special cases only. If such information needs to be added to the map file, we have to thing of another way to do this. (A very simple idea to describe registers which can be reached through different channels, i.e. in the end through different addresses, would be to specify the same register multiple times. This would be an extension, too, but it will not destroy information.)


> The alternative that is being discussed is to keep magic numbers and have meta information in the map files what those magic numbers mean.

This will be very complicated to achieve. These magic numbers are in fact currently not magic numbers, they are addresses. I think you are seeing freedom which is currently not existing, because we already attach too much meaning to this number on the level of the generic NumericAddressedBackend. Still, if we need to add information, why don't we add new fields in the map file instead of abusing existing ones?


> In addition the firmware group wants the map file to work with different transport mechanism (PCIe if you are on a CPU and AXI if you have SytemOnChip). Here some meta information like additional offsets in the AXI address space have to be transported to the backend. This is the type of information that is has to go to the map file, because it depends on the bit-file.

I would need to know more details. My feeling is that this idea somewhat is on the wrong abstraction level. Unless we need an additional level of hierarchy in our address space, the map file does not need to represent certain internal firmware structure.

killenb:
> Again: The bar number must not be abused to encode something which is logically not part of the address. 

We are already doing that. Bar 13 already now does not only contain the information that the ADDRESS column is in a separate virtual address space, but also to use a different transport mechanism (DMA instead of Pcie I/O space). If you really want to separate it we would need a TransportType field, and an index filed for different address spaces with the same transport type.
Lukasz also proposed to add more fields, but this just does not scale. If you want to introduce more and more hierarchy levels it would mean more and more fields, most of them not needed in may cases.

Of course the bar identifier is part of the address, but of the numerical part of the address. You cannot combine addresses from different bars, and addresses from bar 13 are not larger or higher or whatever than addresses from bar 1. They just have nothing to do with each other. 

>> ..keep magic numbers and have meta information in the map files..

> This will be very complicated to achieve.

Not necessarily. A simple meta info entry like `@ADDRESS_SPACE_INFO 7=pcie:0,dma:3` could provide the info to the backend that it can either use pcie BAR 0 or DMA channel 3 to access the register for map file entries with address space ("bar") 7. And if the size in the read() function is larger than 512 bytes, the backed does DMA instead of PCIE i/o. (the xdma driver has only one PCIE bar, but 4 DMA channels). The exact interpretation of the meta info is the part that depends on the backend, and the 512 bytes could be a parameter in the CDD.

>  These magic numbers are in fact currently not magic numbers, they are addresses.

They are unique identifiers which serve as part of the address. But they are not numeric addresses as in 0x400 + 16 bytes = 0x410.

> But the moment these address spaces have an overlap, the NumericAddressedBackend may be confused because it does not know this. 

Exactly. Addresses from different address spaces logically cannot be combined. If we want to play the trick that under the hood different channels can be used, we have to describe it within one register line

> A very simple idea to describe registers which can be reached through different channels, i.e. in the end through different addresses, would be to specify the same register multiple times.

We would have to use special names and combine them under the hood. In the register catalogue there must only be one entry per register. Otherwise it breaks the abstraction. You don't want WORD_TEMPERATURE_VIA_PCIE and WORD_TEMPERATURE_VIA_DMA in your register catalogue, and the application is deciding which one to use when just to do the same thing: read the temperature

> Unless we need an additional level of hierarchy in our address space, the map file does not need to represent certain internal firmware structure.

It probably is more than one hierarchy level which cannot be described at the moment. But it is exactly that kind of information that has to be in the map file: Something that the user of DeviceAccess should not have to care about, but that can change from bit file to bit file. It is the same kind of information as the particular address of a certain register within a bar.

As I said: describing that a single register that can be accessed through different channels is a new concept, and we are just getting the discussion started how to extend the map file to do that. But that is not the primary topic of this ticket.

mhier:
> > Again: The bar number must not be abused to encode something which is logically not part of the address.
> 
> We are already doing that. Bar 13 already now does not only contain the information that the ADDRESS column is in a separate virtual address space, but also to use a different transport mechanism (DMA instead of Pcie I/O space). If you really want to separate it we would need a TransportType field, and an index filed for different address spaces with the same transport type.

Since the Bar 13 refers to a separate address space, this is correct and not abusing anything. We do not care here about details on the transport layer.


> Lukasz also proposed to add more fields, but this just does not scale. If you want to introduce more and more hierarchy levels it would mean more and more fields, most of them not needed in may cases.

Well, if more information is needed, than we need to add more fields. No way around it. Even if we would move to a different format (e.g. XML) we still would have then optional fields - this is no different to adding more fields in the current map file format and define a special value to be placed when this field is not used.


> Of course the bar identifier is part of the address, but of the numerical part of the address. You cannot combine addresses from different bars, and addresses from bar 13 are not larger or higher or whatever than addresses from bar 1. They just have nothing to do with each other.

Correct. This contradicts with using special bar numbers which mean something like "this is a register in bar 1 which can also be transported via dma". In that case there would be a hidden relation between this special bar number, bar 1 and bar 13. The generic NumericAddressedBackend would not know about this relation, hence stuff would break.


> Not necessarily. A simple meta info entry like `@ADDRESS_SPACE_INFO 7=pcie:0,dma:3` could provide the info to the backend that it can either use pcie BAR 0 or DMA channel 3 to access the register for map file entries with address space ("bar") 7. And if the size in the read() function is larger than 512 bytes, the backed does DMA instead of PCIE i/o. (the xdma driver has only one PCIE bar, but 4 DMA channels). The exact interpretation of the meta info is the part that depends on the backend, and the 512 bytes could be a parameter in the CDD.

This logic would then need to go into the generic NumericAddressedBackend!


> > These magic numbers are in fact currently not magic numbers, they are addresses.
> 
> They are unique identifiers which serve as part of the address. But they are not numeric addresses as in 0x400 + 16 bytes = 0x410.

As I said, the complete numeric address is composed of two fields. This kind of implies that the higher-hierarchy field is not a byte address. Yet it is an address, and we assume the address spaces do not overlap.


> > But the moment these address spaces have an overlap, the NumericAddressedBackend may be confused because it does not know this.
> 
> Exactly. Addresses from different address spaces logically cannot be combined. If we want to play the trick that under the hood different channels can be used, we have to describe it within one register line

There are probably many ways how to describe it. The important part is that the generic NumericAddressedBackend must be aware of this. It cannot be done by just reinterpreting the bar number as something different in the backend implementation itself.

I would find it much more logical to use multiple register lines instead, one line for each channel.

> We would have to use special names and combine them under the hood. In the register catalogue there must only be one entry per register. Otherwise it breaks the abstraction. You don't want WORD_TEMPERATURE_VIA_PCIE and WORD_TEMPERATURE_VIA_DMA in your register catalogue, and the application is deciding which one to use when just to do the same thing: read the temperature

You are overthinking this. No need for special names. Just multiple entries with the exact same name but different addresses (bar + byte offset). We could teach the NumericAddressedBackend to be aware that these are the pointing to the same register.


> It probably is more than one hierarchy level which cannot be described at the moment. But it is exactly that kind of information that has to be in the map file: Something that the user of DeviceAccess should not have to care about, but that can change from bit file to bit file. It is the same kind of information as the particular address of a certain register within a bar.

So far I have not yet seen any need for an additional level of hierarchy. (Well, maybe we want bit-level access, but that's a different story.)


> As I said: describing that a single register that can be accessed through different channels is a new concept, and we are just getting the discussion started how to extend the map file to do that. But that is not the primary topic of this ticket.

Yes, but abusing the bar number for this feature will break things. Anyway, this feature is probably 10x more complicated than everyone thinks. The decision which channel to use will be highly non-trivial in the end. I seriously doubt there will be an actual benefit from this feature, because in the end it is better to manually decide case by case which channel to use. Just to give you an idea what we are talking about: Imagine 2 registers which can do both DMA and PIO transfers. Each of them is too small to do an efficient DMA transfer when used alone. Now you place both into a TransferGroup and the transfer can be merged. As a consequence a DMA transfer might be more efficient, so the accessors have to be changed from PIO to DMA. Fun fact: transfer merging is decided on the level of the generic NumericAddressedBackend, while the decision to do PIO or DMA can only be done by the backend itself. To make things even worse, what should now happen if a third register is added to the TransferGroup which can do only PIO? Is it more efficient to merge the transfers and move everything back to PIO, or better to keep the transfer separate and use DMA where possible?

TL;DR: we do not want overlapping address spaces. Things might be fine if entire address spaces can do both PIO or DMA, but the DMA channel and PIO bar must not contain other registers which cannot do both... If this is the case, some additional metadata info might work (but there is no need to invent new bar numbers then, just keep the true bar number and add the information which DMA channel can be used to access them too).


Sorry for abusing this thread, but afaik there is no better place to discuss this right now. I just want to make sure we are not breaking the framework in a subtle way by changing the meaning of the bar number for new use cases.

killenb:
I get the impression that we both mean the same, but say it with different words and don't understand each other.

> About naming: What about renaming "bar" and "address" into "super-address" and "sub-address"?

We definitely have to do something about the naming. This is part that causes most confusion and misunderstanding. I like to think of bar as "address space", and "address" as ..., well, address in that address space. For me it makes intuitively clear that addresses in different bars have nothing to do with each other. However, the complete address which is unique consists of both, and to call only the numeric part an address is not good. I have no idea how to transport both facts in the same wording.

### I will try to start from scratch to explain what I want to say.

The part where I think we both agree:

* For the NumericAddressedBackend base class and its accessors the "bar" is number that distinguishes address spaces.
* Only addresses in the same bar can be combined. If they are in different bars they can never be combined.
* Addresses in the same bar can always be combined.

This is the part that is implemented in the accessors which allows to combine adjacent registers to use the same, larger, low level transfer element.

* The "bar" number is never interpreted by the NumericAddressedBackend and its accessors. It is the first part of the unique register address, and the only check that is done is for equality to see if registers can be combined.

This is the part that I do not want to touch. I completely agree that it would be complicated to teach the NumericAddressedBackend to combine anything else than neighbouring addresses in the same bar. As you say it has the potential to break abstraction, and probably will if we try it.

* The transport layer implementation completely depends on the individual backend
* The transport layer implementation can be different for different bars (e.g. BAR 13 for the pcie backend is DMA).

### The part that needs to be extended

* The xdma interface has multiple configuration options. We need a ways to get this configuration to the transport layer implementation

Usually this is done with the CDD. However, parts of this information only depend on the bit file, and can potentially change from compilation to compilation. This is the kind of information that traditionally is in the map file.

**Example:**

Registers in BAR 7 can be accessed though 3 different channels:
1. memory mapped when connected via PCIe.
2. DMA when connected via PCIe
3. AXI bus in SystemOnChip systems. In this case a certain offset needs to be known. This offset should be in the map file.

* With this additional information the transport layer can chose a different transport channel, even within one bar.

This is a pure implementation detail of the individual backend, and the decision what to use can only be done in this layer. There is not way to do this kind of decision in the NumericAdddressedBackend or higher without breaking the abstraction.

patrislav1:
> Registers in BAR 7 can be accessed though 3 different channels:
> 
>     1. memory mapped when connected via PCIe.
> 
>     2. DMA when connected via PCIe

I thought we follow the existing convention and use BAR 13 for DMA and lower BARs for PIO access?

>     3. AXI bus in SystemOnChip systems. In this case a certain offset needs to be known. This offset should be in the map file.

In that case, there is a `uio` device file for _each_ FPGA IP which shows up with a name that's defined in the Linux device tree (`.dts` / `.dtb`) like this:
```
hier_daq_arm_axi_traffic_gen_0@88820000 {
    compatible = "generic-uio";
    reg = <0x00 0x88820000 0x00 0x10000>;
};

hier_daq_pcie_axi_bram_ctrl_0@88920000 {
    compatible = "generic-uio";
    reg = <0x00 0x88920000 0x00 0x2000>;
};

hier_daq_pcie_axi_dma_0@88910000 {
    compatible = "generic-uio";
    reg = <0x00 0x88910000 0x00 0x1000>;
};
```
With above .dts code (from a ZUP example project), the kernel creates device files `/dev/uio/hier_daq_arm_axi_traffic_gen_0`, `/dev/uio/hier_daq_pcie_axi_bram_ctrl_0`, `/dev/uio/hier_daq_pcie_axi_dma_0`.
In this case a future AXI backend will _not_ need the start addresses of the IP cores, as the kernel will use them implicitly. It will however need the register offsets relative to the start of each IP (just as the pcie backend), and the device file name for each IP core (_instead of_ the start address).

This means, if we want to share/reuse map files between both PCIe and AXI transport layers, that map file will need to describe each IP with _both_ a start address (for PCIe) _and_ a device name (for AXI).
If that's not possible, the device name will need to be in the CDD / dmap file (as @mhier suggested) and then each single FPGA IP will need to have a map file of its own.

> We definitely have to do something about the naming. This is part that causes most confusion and misunderstanding. I like to think of bar as "address space"

Why don't we just call it `addressSpace` or similar? It seems to sufficiently describe what it's doing, and fits the PCIe `BAR` context as well as the more generic one in XDMA and others.

mhier:
> I thought we follow the existing convention and use BAR 13 for DMA and lower BARs for PIO access?

I think this was merely an example what future special bar numbers could mean.


> This means, if we want to share/reuse map files between both PCIe and AXI transport layers, that map file will need to describe each IP with _both_ a start address (for PCIe) _and_ a device name (for AXI).

This can be solved using the subdevice backend already now. The "IP core" (which is a misleading name, too) has a map file with addresses relative to the start of the "IP core". This is directly used for the AXI interface. For PCIe we need another map file with each one AREA for each "IP core". The subdevice backend is then instructed to create a new device for this AREA using the map file for the "IP core".


> If that's not possible, the device name will need to be in the CDD / dmap file (as @mhier suggested) and then each single FPGA IP will need to have a map file of its own.

This solution will still require one CDD entry per "IP core" for both transport layers. In any case I think it is a good idea to have the device structure independent of the transport layer, i.e. if AXI requires us to make one device per "IP core", we should aim for the same for PCIe (extra devices to organise this like the one holding all the AREAS don't count, as they are not directly used by the application).


> Why don't we just call it `addressSpace` or similar? It seems to sufficiently describe what it's doing, and fits the PCIe `BAR` context as well as the more generic one in XDMA and others.

Because then the `addressSpace` is part of the address of our address space... ;-) IMO the term address space should refer to the entire address space of the device, which includes all (valid) "bar numbers" (or whatever we may call them). Everything else will be highly misleading, since the NumericAddressedBackend is simply built that way: It assumes every register has one unique (sic!) address, and this address has the format `uint8_t.uint32_t`.

PR #210, opened 2021-06-25
SubdeviceBackend 2regs/3regs mode: add support for data areas.

### LINKED ISSUE #177: SubdeviceBackend: 2regs/3regs mode with long data register
One use case for the SubdeviceBackend is currently not covered: The llrfctrl server is supposed to write (for the Klystron predistorter) several tables to the firmware. To save address space on the firmware side, it was decided to use only one AREA register for the tables and multiplex it through a selector bit field. This is in principle similar to the use case of the attenuators, which are single-word values multiplexed with a selector bit field.

The difference comes from then length of the data register. Currently, it only properly works for single-word data registers. This needs to be extended for multi-word data registers.

Terminology used in this ticket:
- _data register_ / _address register_: registers in the target device as specified via the data and address CDD parameter
- _subdevice register"_ virtual register which gets unfolded by the SubdeviceBackend and is presented to the application as a normal register.

The length of the data register determines the granularity of the address space defined by the _address register_. It is important that the start addresses of the _subdevice registers_ specified in the map file are written unaltered to the address register in the target device. This must hold even if only parts of the _subdevice register_ is accessed. Partial access must instead result in partial access of the _data register_.

The current implementation wrongly uses only the first word of the _data register_ and will increment the value of the _address register_ for the subsequent word accesses.

Note: The SubdeviceBackend is not a NumericAddressedBackend, for good reasons. Access must always be directed to a single _subdevice register_ - it cannot happen that access is spanning across multiple _subdevice registers_, hence the _address registrer_ must never be incremented during a single access. It must **not** be possible to merge multiple operations into a single one via a TransferGroup (as it is possible for the NumericAddressedBackends). Merging simply would not work because it is not clear by how much the _address register_ would need to be changed. The _address register_ can be even a bit mask, resulting in a non-linear address space (all addresses are a power of 2). Hence we need to treat the address space as non-continuous.

mhier:
This fixes #177 (on github)

Note: This changes existing behaviour in one aspect: previously, the
address written to the address register was incremented in steps of 4
for subsequent low-level transfers of the same high-level register (like
a byte address). This does not make much sense, and likely no use case
relied on this behaviour. It should be noted though that this behaviour
was tested. The tests have been changed accordingly (and extended with
the new functionality).

### DIFF
--- a/device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
+++ b/device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
@@ -10,7 +10,7 @@ namespace ChimeraTK {
       boost::shared_ptr<NDRegisterAccessor<int32_t>> accData, boost::shared_ptr<NDRegisterAccessor<int32_t>> accStatus,
       size_t byteOffset, size_t numberOfWords)
   : NDRegisterAccessor<int32_t>(registerPathName, {AccessMode::raw}), _backend(backend), _accAddress(accAddress),
-    _accData(accData), _accStatus(accStatus), _byteOffset(byteOffset), _numberOfWords(numberOfWords) {
+    _accData(accData), _accStatus(accStatus), _startAddress(byteOffset), _numberOfWords(numberOfWords) {
     NDRegisterAccessor<int32_t>::buffer_2D.resize(1);
     NDRegisterAccessor<int32_t>::buffer_2D[0].resize(numberOfWords);
     _buffer.resize(numberOfWords);
@@ -26,9 +26,11 @@ namespace ChimeraTK {
 
   bool SubdeviceRegisterAccessor::doWriteTransfer(ChimeraTK::VersionNumber) {
     std::lock_guard<decltype(_backend->mutex)> lockGuard(_backend->mutex);
+    // This is "_numberOfWords / _accData->getNumberOfSamples()" rounded up:
+    auto nTransfers = (_numberOfWords + _accData->getNumberOfSamples() - 1) / _accData->getNumberOfSamples();
     try {
       size_t idx = 0;
-      for(size_t adr = _byteOffset; adr < _byteOffset + 4 * _numberOfWords; adr += 4) {
+      for(size_t adr = _startAddress; adr < _startAddress + nTransfers; ++adr) {
         if(_backend->type == SubdeviceBackend::Type::threeRegisters) {
           size_t retry = 0;
           size_t max_retry = _backend->timeout * 1000 / _backend->sleepTime;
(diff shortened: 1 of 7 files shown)


device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
```diff
@@ -26,9 +26,11 @@ namespace ChimeraTK {
 
   bool SubdeviceRegisterAccessor::doWriteTransfer(ChimeraTK::VersionNumber) {
     std::lock_guard<decltype(_backend->mutex)> lockGuard(_backend->mutex);
+    // This is "_numberOfWords / _accData->getNumberOfSamples()" rounded up:
+    auto nTransfers = (_numberOfWords + _accData->getNumberOfSamples() - 1) / _accData->getNumberOfSamples();
     try {
       size_t idx = 0;
-      for(size_t adr = _byteOffset; adr < _byteOffset + 4 * _numberOfWords; adr += 4) {
+      for(size_t adr = _startAddress; adr < _startAddress + nTransfers; ++adr) {
```
killenb, line 33: Having more than one transfer and increasing the address contradicts the comment in the ticket that address spaces can be non-contiguous (for instance if they are bit patterns)

device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
```diff
@@ -26,9 +26,11 @@ namespace ChimeraTK {
...
         if(_backend->type == SubdeviceBackend::Type::threeRegisters) {
```
killenb, line 34: I know of at least two places with a 1-D register (four 32-bit words) where I have to wait for the busy register to clear after each word that is written. Previously only one word per address was written, so it does not make a difference where the code is waiting for the flag to clear.

I am not sure how the firmware is implemented in this case, and if it is the same in all cases, but there are two possibilities where the code might have to wait:
* after each word
* only after writing everything that belongs to one address (this implementation)

Maybe needs clarification with the FW team.

device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
mhier: True, but I did not want to break this existing feature at this time. Since this depends only on information present in the map file, one simply needs to pay attention to define registers which are not longer than the target data register when the address space is non-contiguous.

device_backends/Subdevice/src/SubdeviceRegisterAccessor.cc
mhier: That would be a different use case. In my understanding, the case you are mentioning does not have an address register at all, because the address unfolding is already done in firmware. That would hence be more like an extension to the area mode of the SubdeviceBackend (this feature is only about the 2regs/3regs mode). Needing to wait after each word of the data register in 2/3regs mode sounds very weird, I don't think that such a case exists (clearly not the use case currently at hand). If we need this in future, we can still extend by adding a new parameter.

PR #211, opened 2021-07-01
Add XDMA backend

patrislav1:


device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -0,0 +1,126 @@
...
+    return true;
+  }
+
+  XdmaIntfAbstract* XdmaBackend::_intfFromBar(uint64_t bar) {
+    if(bar == 0) {
+      return _ctrlIntf ? dynamic_cast<XdmaIntfAbstract*>(&_ctrlIntf.value()) : nullptr;
+    }
+    // 13 is magic value for DMA channel (by convention)
+    // We provide N DMA channels starting from there
+    const ssize_t dmaChIdx = static_cast<ssize_t>(bar) - 13;
```
killenb, line 60: Why do you do all the gymnastics with converting to signed and then having negative values?
You could just check that  bar is >= 13 and then keep using unsigned values. Avoids a couple of static_casts.

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -0,0 +1,126 @@
...
+    if(dmaChIdx < 0 || dmaChIdx >= static_cast<ssize_t>(_dmaChannels.size())) {
+      auto exc_string =
+          std::string{"Couldn't find XDMA channel for BAR value "} + std::to_string(static_cast<int>(bar));
```
killenb, line 63: Opps, this static cast was overlooked in the merge with master. When bar was still uint8_t it would have been treated as a character, so the cast was necessary. With uin64_t this is not needed any more. My bad.

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -0,0 +1,126 @@
...
+#endif
+    auto intf = _intfFromBar(bar);
+    intf->write(address, data, sizeInBytes);
+#ifdef _DEBUGDUMP
+    dump(data, sizeInBytes);
+#endif
+  }
+
+  std::string XdmaBackend::readDeviceInfo() {
+    // TODO implement me
```
killenb, line 111: At least return that this is the XMDA backend. If you add the _devicePath it's basically complete.

device_backends/xdma/udev_rules/xdma-udev-command.sh
```diff
@@ -0,0 +1,45 @@
+#!/ bin / bash
```
killenb, line 1: Move this file to the driver package.

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -0,0 +1,126 @@
...
+    _dmaChannels.clear();
+  }
+
+  bool XdmaBackend::isOpen() { return !!_ctrlIntf; }
+
+  bool XdmaBackend::isFunctional() const {
+    if(!_ctrlIntf || _hasActiveException) {
+      return false;
+    }
+    // TODO implement me
```
killenb, line 50: Seems sufficient to me. The next open will detect this and try a re-connect.

killenb: Well, yes. You could actively try to read something to see if the device really is still working.

device_backends/xdma/udev_rules/60-xdma.rules
```diff
@@ -0,0 +1,12 @@
+#The kernel creates the unique device nodes under / dev.These cannot be changed or renamed.
```
killenb, line 1: Move this file to the driver package

killenb (changes requested):

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -57,13 +56,13 @@ namespace ChimeraTK {
     }
     // 13 is magic value for DMA channel (by convention)
     // We provide N DMA channels starting from there
-    const ssize_t dmaChIdx = static_cast<ssize_t>(bar) - 13;
-    if(dmaChIdx < 0 || dmaChIdx >= static_cast<ssize_t>(_dmaChannels.size())) {
-      auto exc_string =
-          std::string{"Couldn't find XDMA channel for BAR value "} + std::to_string(static_cast<int>(bar));
-      throw ChimeraTK::logic_error(exc_string);
+    if(bar >= 13) {
```
killenb, line 59: Sehr schöne Lösung.

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -108,8 +107,12 @@ namespace ChimeraTK {
   }
 
   std::string XdmaBackend::readDeviceInfo() {
-    // TODO implement me
-    return "";
+    if(!isOpen()) throw ChimeraTK::logic_error("Device not opened.");
```
killenb, line 110: The device info usually is read before opening the device. I'll quickly change that to print 
`number of DMA channels = unkown (device closed)` instead of throwing.

patrislav1: That's interesting, as I copied it from the [pcie backend implementation](https://github.com/ChimeraTK/DeviceAccess/blob/9e585a67a8222c856f1ef4aedceeda86b469df1f/device_backends/pcie/src/PcieBackend.cc#L277).

killenb: Seems you found a bug in the PCIe backend. Plus it seems this function is not used very much, or at least not called on a closed device as I expected.

killenb (approved):

patrislav1 (commented):
typo: uknown/unknown

PR #213, opened 2021-08-16
Added docu for the design of AsyncNDRegisterAcessor ... and the asynchron…

killenb:
...and the asynchronous implementation in NumericAddressedBackend

Please read through the documentation page and give feedback


doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+\subsection design_async_map_file Asynchronous registers in the map file
+
+<pre>
+# name            nElements  address  nBytes  bar  width  bitInterpretation signed accessMode
+APP.0.DATA_READY          0        0       0    0      0                  0      0 INTERRUPT1:5
+APP.0.PUSH_DATA           4       12      16    1     32                  8      1 INTERRUPT2:3
+</pre>
+
+The map file has been extended by the access mode INTERRUPTx:y. Before, only RW and RO were allowed. x and y are the number of the interrupt controller and the interrupt number, respectively. Registers with access mode INTERRUPT will implicitly be treaded as read-only.
```
phako, line 55: treaded -> treated

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
+namespace ChimeraTK {
+/**
+\page design_AsyncNDRegisterAccessor_NumericAddressed Design: AsyncNDRegisterAccessor and the asynchronous implelentation in the NumericAddressedBackend
```
phako, line 3: implelentation -> implementation

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+\section  design_AsyncNDRegisterAccessor The AsyncNDRegisterAccessor
+
+The AsyncNDRegisterAccessor is an extension of the NDRegisterAccessor for push-type data.
+
+The NDRegisterAccessor itself extends the TransferElement with a user-type buffer. The TransferElement does not have a user type buffer and serves as a base class such that implementations with different user types can be stored together in containers. The TransferElement also contais a notification queue to implement the asynchronous read functionality. This notification queue is of type void (notifications only, no data is transported) because the data type is not known. The AsyncNDRegisterAccessor now povides the data transport queue which is needed for asynchronous accessors with data. As this queue is comparatively resource-intensive (multiple instances of the user buffer object pre-allocated in each queue) and not needed for synchronous data transfer, it is not directly added in the NDRegisterAccessor but in a second class.
+
+The data transport queue is implmented as continuation of the notification queue, such that the read operations of the notification queue trigger the filling of the user buffer.
```
phako, line 11: implmented -> implemented

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
+namespace ChimeraTK {
+/**
+\page design_AsyncNDRegisterAccessor_NumericAddressed Design: AsyncNDRegisterAccessor and the asynchronous implelentation in the NumericAddressedBackend
+
+\section  design_AsyncNDRegisterAccessor The AsyncNDRegisterAccessor
+
+The AsyncNDRegisterAccessor is an extension of the NDRegisterAccessor for push-type data.
+
+The NDRegisterAccessor itself extends the TransferElement with a user-type buffer. The TransferElement does not have a user type buffer and serves as a base class such that implementations with different user types can be stored together in containers. The TransferElement also contais a notification queue to implement the asynchronous read functionality. This notification queue is of type void (notifications only, no data is transported) because the data type is not known. The AsyncNDRegisterAccessor now povides the data transport queue which is needed for asynchronous accessors with data. As this queue is comparatively resource-intensive (multiple instances of the user buffer object pre-allocated in each queue) and not needed for synchronous data transfer, it is not directly added in the NDRegisterAccessor but in a second class.
```
phako, line 9: I suppose this is meant as a motivation why there is the need for an additional class? I would mention that at the beginning, otherwise the explanation of the TransferElement internals sounds a bit random here.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+The only function the backend implementor really has to deal with is NumericAddressedInterruptDispatcher::trigger(). All other functionaltiy for subscription, activation/deactivation and exception handling is already handled in the NumericAddressedBackend.
+
+There is one NumericAddressedInterruptDispatcher per interrupt controller number and interrupt number, stored in the variable NumericAddressedBackend::_interruptDispatchers. The interrupt handling thread in the backend implementation will just call
+
+<pre>
+  _interruptDispatcher.at({interruptControllerNumber, interruptNumber})->trigger();
+</pre>
+
+Notice that you should access the map with at() and not through the [ ] operator, so you don't accidentally add empty dispatchers to the map.
```
phako, line 41: If this is the only interface to the _interruptDispatcher from implementing classes, why don't we make it private and add a `get()` function for the `(controller, interrupt)` pair?

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+The interrupt handler thread should have a "lazy start", which means that it should only be started once accessors to the particular interrupt have been requested. To achieve this, the backend should override NumericAddressedBackend::startInterruptHandlingThread(unsigned int interruptControllerNumber, unsigned int interruptNumber). This function is called each time an asynchronous accessor which is associated to that interrupt is created. The implementing code has to make sure that the thead is only started once. The thread in not terminated automatically. The backend has to take care to do this when the backend is closed and/or destructed.
```
phako, line 42: thead -> thread

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+Internally an object called NumericAddressedAsyncVariable (or to be more precie a shared pointer to the typed implementation NumericAddressedAsyncVariableImpl) is stored for each required accessor type. This contains one synchronous read accessor together with all its subscribers. Like this the hardware access to the devide and the possibly expensive postRead() action (de-mupltiplexing and numeric conversion) are only done once, before the data is pushed into the data transport queues of the subscribers.
```
phako, line 44: - precie ->precise
- devide -> device
-

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+There can be multiple NumericAddressedAsyncVariables for one register. For example one for raw access, one with data conversion and a thrid one for only the first 10 elements in a large 1D register. All this is happening automatically in the NumericAddressedBackend and the developer of the individual backend does not have to care about it.
```
phako, line 45: thrid -> third

phako (changes requested):

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
mhier: This is the wrong direction. The notification queue is a continuation of the data transport queue.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
mhier: The title refers to the numeric addressed backend, but it seems at least a great part of this document is independent it. I would remove the reference to the NA backend from the title.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+The data transport queue is implmented as continuation of the notification queue, such that the read operations of the notification queue trigger the filling of the user buffer.
+
+The AsyncNDRegisterAccessor is ready to use. It provides a generic asynchronous implementation.
+Each backend has to store a list of accessors and call AsyncNDRegisterAccessor::activate(), AsyncNDRegisterAccessor::deactivate(), AsyncNDRegisterAccessor::sendDestructively() and AsyncNDRegisterAccessor::sendException() appropriately. As the way the data containers are organised and the asynchronous mechanism is implemented stongly depend on the backend, the implementation of these are left to the specific backend.
```
mhier, line 14: Does the AsyncNDRegisterAccessor hold a shared_ptr to the backend? In that case the backend must use a weak_ptr. This should be clarified either way. (You mention this below in context of the manager, but here it is the same, right?)

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+However, there is a common interface implemented to allow the clean subscription/unsubscription from the backend. The AsyncNDRegisterAccessor is templated to an AsyncAccessorManager and an AsyncAccessorID, which are both handed over in the constructor. In its destructor the AsyncNDRegisterAccessor is calling
+<pre>
+void AsyncAccessorManager::unsubscribe(AsyncAccessorID id);
+</pre>
+This function has to be implemented by the manager. From the ID it knows which subscriber is to be removed. For this mechanism to work, the AsyncAccessorManager must store boost::weak_pointers and lock them when the instance is being accessed. Otherwise the one instance held within the manager prevents the destructor from being called.
```
mhier, line 19: The structure is a but unclear to me. It looks like one has to use a manager (because the accessor is templated to it). Before you state that the functions `activate()` etc. need to be called by the backend. This is confusing. They should be called by the manager, right?

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+\section  design_async_NumericAddressedBackend Asynchronous accessors in the NumericAddressedBackend
+
+As the NumericAddressedBackend base class already provides the creation of the synchronous NumericAddressedBackendRegisterAccessors, it also provides the subscription mechanism for the asynchronous accessors. The NumericAddressedBackend is based on a map file which allows to describe registers with interrupts. In the interface to the hardware, all interrupts are void. The map file then descibes which of the synchronous registers are read out and pushed when the according interrupt is triggered. This mechanism is implemented in the NumericAddressedBackend base class and the protocol implementing backends only have to notify the NumericAddressedInterruptDispatcher that the interrupt has happened.
+
+\subsection  async_NumericAddressedBackend_interface Interface for backend implementations
+
+The NumericAddressedInterruptDispatcher serves two purposes:
+\li It serves as a subscription manager for the AsyncNDRegisterAccessor
+\li It provides functions which act on all subsribers.
```
mhier, line 29: Doesn't it also handle the readouts through the synchronous accessors and passing on the (copied) data to the async accessors?

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
mhier: Also I am wondering if we could (maybe later) split a generic portion off the NumericAddressedInterruptDispatcher into an AsyncAccessorManager base class, which would not be mandatory to use but likely used by most backends. This could deal with maintaining the subscriptions and copying the incoming data and passing it to all subscribers, so the backend has to call merely one function when new data arrives.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
mhier: "The thread in not" -> "The thread is not"

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
mhier: Sounds like duplicated functionality with the TransferGroup.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,60 @@
...
+
+Internally an object called NumericAddressedAsyncVariable (or to be more precie a shared pointer to the typed implementation NumericAddressedAsyncVariableImpl) is stored for each required accessor type. This contains one synchronous read accessor together with all its subscribers. Like this the hardware access to the devide and the possibly expensive postRead() action (de-mupltiplexing and numeric conversion) are only done once, before the data is pushed into the data transport queues of the subscribers.
+There can be multiple NumericAddressedAsyncVariables for one register. For example one for raw access, one with data conversion and a thrid one for only the first 10 elements in a large 1D register. All this is happening automatically in the NumericAddressedBackend and the developer of the individual backend does not have to care about it.
+
+\subsection design_async_map_file Asynchronous registers in the map file
```
mhier, line 47: Maybe move this subsection up, since you are referring to the interrupt controller and interrupt number concepts in the previous subsection already.

mhier (changes requested):

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Good idea. Probably even better: Have a function `void dispatchInterrupt(controller, interrupt)` and don't let the implementing code fiddle with the dispatcher at all. I will check that.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Hm, yes. Now I remember that I wanted to run a spell checker before the commit.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Yes, the manager is the part of the backend which is holding the weak pointers.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: I had the same idea. But the way it is done in the NumericAddressedBackend is already very special. The actual list of weak pointers is inside of a helper object AsyncVariable, which contains all the subscribers for one particular accessor, together with the according synchronous accessor to read the data. That is the first list which changes with a subscription/unsubscription.
The dispatcher then holds a list of those AsyncVariables, which can also grow/shrink. All those lists have to be protected with a lock against the dispatcher thread.
One would only benefit if the subscribe/unsubscribe would be in the generic manager, but I did not see how to fit this without the special behaviour of one async variable which feeds a sub-set. The Doocs-bakend would have a completely different mechanism.

killenb: ~~The Doocs-bakend would have a completely different mechanism.~~
Scratch that. It's exactly the same. Multiple subscribers to the same variable must be handled, and there is a list of variables.

killenb: I will see if I can already separate that in a clever way. But the current implementation has a lock. I don't know if this is acceptable for a generic solution.

For the AsyncNDRegisterAccessor it does not make a difference. But having a base class which does the fiddly stuff and has some rules is a good idea

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Yes, that's one of the functions that acts on all subscibers. I guess I should re-phrase that.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Well, the complete second section of the document is only about the NumericAddressedBackend. And this is the bigger part. Only the first section is generic.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: A bit. But the transfer group does a lot of magic to get the pre- and post operations into the right order, re-grouping the transfers under hood etc, All this is not happening. There internally is one complete synchronous transfer, and then the queues are filled. It's more like a fan-out.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+
+One AsyncAccessorManager is used per interrupt that can occur. The AsyncAccessorManager has three main functions:
+
+- It serves as a factory for AsyncNDRegisterAccessors
+- It contains a list of created asynchronous accessors and a subscribe/unsubscribe mechanism
+- It provides functions to act on all asynchronous accessors
+
+\subsection AsyncNDRegisterAccessor_details Design decisions and implementation details
+
+- The data transport queue is of type UserType, so the potentially expensive and slow operations (data readout,
```
mhier, line 27: `The data transport queue is of type UserType` - this is certainly wrong. I guess it is of the type `NDRegisterAccessor<UserType>::Buffer` (if I understand correctly, see above).

mhier: You probably mean, the data transport queue contains data already converted to the UserType, right?

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+  de-multiplexing, data conversion) are happening only once, before the data is distributed to multiple subscribers.
+- The additional send buffer is necessary because data has to be send through the queue in this container, and usually
```
mhier, line 29: I do not understand this point. Is the "buffer" here the type again, or an instance?

If type is meant: The type is merely needed to aggregate the information for the queue.

If instance is meant: I do not understand from this.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+  the data is not provided in this format.
+  Even if the NDRegisterAccessor would provide the user buffer in this format it could not be used here, because it is
+  used by the receiving side which is running in another thread.
+- To unsubscribe, the NDRegisterAccessor is calling AsyncAccessorManager::unsibscribe() with its TransferElementID from
+  the destructor.
+  For the unsubscribe mechanism to work, the AsyncAccessorManager can only hold a weak pointer.
+  If it would hold a shared pointer, the destructor would never be called because this instance survives,
+  even if the user code has deleted all copies of the AsyncNDRegisterAccessor.
+- To manage the weak pointer and to have an entry object for the implementation, the helper class
```
mhier, line 38: The justification for the AsyncVariable is too low level for my taste. It is better to explain what it does rather then why you felt it was necessary to introduce it. I still don't really understand what it does. Does it represent a single AsyncNDRegisterAccessor inside the AsyncAccessorManager? I think, this is all what matters here.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+- AsyncVariable::getNumberOfChannels(), AsyncVariable::getNumberOfSamples(), AsyncVariable::getUnit(),
+  AsyncVariable::getDescription() and AsyncVariable::isWriteable() must be provided by deriving from AsyncVariable or
+  AsyncVariableImpl. The information must directly be provided by the backend.
+  The generating code cannot rely on the catalogue here (which might not even be filled yet when the accessor
+  is being created).\ref asyncAccessor_comment_AsyncVariable "(*2)"
+- The 'virtual function template' AsyncAccessorManager::createAsyncVariable() must return an AsyncVariable
+  implementation with the implemented accessor-describing functions from the previous bullet point, such that the
+  accessor manager can create the AsyncNDRegisterAccessor.
+  When createAsyncVariable() returns the AsyncVariable object, the weak pointer to the AsyncNDRegisterAccessor is not
+  set yet. The reason is that the backend specific code should not have the subscription code, which is the same for
```
mhier, line 66: I think the reasoning here is more distracting than helping. Once the reader has understood enough context to be able to comprehend what you are saying here, the reason already is self-evident.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+  all backends and would only be code duplication. And it is not sufficient to have the AsyncVariable, which only
+  contains the weak pointer. The calling code also needs the shared pointer to the AsyncNDRegisterAccessor.
+  createAsyncVariable() is used internally by the AsyncAccessorManager, which creates the AsyncNDRegisterAccessor
+  and adds it to the AsyncVariable in the subscription process.
+- AsyncAccessorManager::activate() needs to be implemented in the following way:
+  -# Lock the mutex which protects the list of weak pointer, so no entries can be added or removed
```
mhier, line 72: You lost me again here... Where does the mutex live, and where the weak pointer? I though the weak pointer is part of the AsyncVariable? So is this actually the list of AsyncVariables? For the mutex, writing out the class member name would help.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+  
+  Locking of the mutex, generating the version number and calling activate() on each variable could be factored into
+  the manager, with additional virtual functions that are called in between, but this means the container needs to be
+  looped several times and the design gets more complicated and involved,
+  and hence the implementation is left to the backend.
+
+The backend uses the AsyncNDRegisterAccessor by
+- creating one AsyncAccessorManager for each interrupt that can occur (or for each push type variable that is being
+  subscribed from a sender)
+- calling AsyncAccessorManager.subscribe() in getRegisterAccessor(), and handing out the created AsyncNDRegisterAccessor.
```
mhier, line 86: use `::` instead of `.` for the member function name...

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+  createAsyncVariable() is used internally by the AsyncAccessorManager, which creates the AsyncNDRegisterAccessor
+  and adds it to the AsyncVariable in the subscription process.
+- AsyncAccessorManager::activate() needs to be implemented in the following way:
+  -# Lock the mutex which protects the list of weak pointer, so no entries can be added or removed
+  -# Generate a new version  number while holding the lock
+  -# Write the initial value, the according data validity flag and the version number to the _sendBuffer of each
+     AsyncVariable
+  -# Call activate() on all AsyncVariables
+  
+  Locking of the mutex, generating the version number and calling activate() on each variable could be factored into
```
mhier, line 78: move to footnote maybe?

mhier (changes requested):

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
+namespace ChimeraTK {
```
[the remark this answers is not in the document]
killenb: I had the soft wrapping before which ended in very long lines and the diff view in basically any tool being useless. So I 
formated to the 120 characters we also use for source code.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
```diff
@@ -0,0 +1,158 @@
...
+
+\section  design_AsyncNDRegisterAccessor The AsyncNDRegisterAccessor and the AsyncAccessorManager
+
+The AsyncNDRegisterAccessor is the NDRegisterAccessor implementation for push-type data. It is used
+if accessors with AccessMode::wait_for_new_data are requested. See \ref transferElement_B_8
+"Technical specification: TransferElement B.8" for the requirements it is based on.
+
+The AsyncNDRegisterAccessor is a generic implementation for all backends and contains the following components
+
+- an NDRegisterAccessor<UserType>::Buffer which serves as a send buffer and contains the data value
```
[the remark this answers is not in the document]
killenb: NDRegisterAccessor<UserType>::Buffer is a type, and the async has an instance of this which is called _sendBuffer.

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: With "send buffer" the instance is meant. You have to prepare an instance of the object that is being pushed into the queue because it is swapped with an instance inside the queue when pushing.
When popping the queue this is not necessary if you have set up a continuation. The function that is called there is getting a reference of the received element, and you can extract the data and put it into buffer_2D etc. For sending this is not possible, so here you need an instance of the same type as it is running in the queue, prepare it and then call push.

How do I phrase that?

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Yes, in the new design it reprecents one AsyncNDRegisterAccessor

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: ah yes. I see the confusion. I will have a look at the class description. I think there it is described better.

The mutex protects the list of AsnycVariables in the manager. Whenever you want to access one of those variables, you have to hold the mutex because an un-subscribe could delete it.
So the mutex serves two purposes.
1. protects the existence of the variable.
2. As you have to hold it anyway, it also protects the contents (sending expections, enabling/diabling and sending data do not all come from the same thead).

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: good idea, better for the flow

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: I tried to shorten it. Section still is not very catchy

doc/design_AsyncNDRegisterAcessor_and_NumericAddressedBackend.dox
killenb: Wow, this was really confusing and wrong. There is no list of weak pointers, there is a list (actually a map) of AsyncVariables.

killenb: OK, I think I got this one now.

PR #218, opened 2022-01-17
Fix ScalarRegisterAccessors<Boolean> not implicitly converting to bool.

mhier:
**Background:**

Implicit conversions of ```ScalarRegisterAccessors``` to their ```UserType``` are
frequently used. In case of ```ChimeraTK::Boolean```, often a conversion to
the plain ```bool``` type is required instead, e.g.:

```
ScalarRegisterAccessor<Boolean> myAccessor;
// obtain accessor and read value.
if(myAccessor) doSomething();
```

Without this patch, one had to write instead:
```
if(bool(myAccessor))  doSomething();
```
or
```
if(Boolean(myAccessor)) doSomething();
```

**Implementation:**

This is implemented by adding a template specialisation of
```ChimeraTK::Boolean``` which has implicit conversion to ```bool```. To avoid full
code duplication of the entire class, a second template argument ```TAG``` is
introduced, which allows the template specialisation to inherit
everything else from the generic implementation. This should not have
any negative side effects (apart from maybe cluttering some compiler
error messages with the additional template argument).

**Caveat:**

The change may require a code change of other libraries/application in
some less frequently used cases. This code no longer works:
```
ScalarRegisterAccessor<Boolean> acc1, acc2;
// obtain accessors and read acc1.
acc2 = Boolean(acc1);
```
because the conversion to ```ChimeraTK::Boolean``` is now ambigous. Instead
one has to write:
```
acc2 = bool(acc1);
```
This is mainly problematic in places where template code assigns one
accessor value to another accessor, so an exception has to be
introduced, e.g.:
```
template<typename UserType>
void myFn(ScalarRegisterAccessor<UserType> a,
          ScalarRegisterAccessor<UserType> b) {
  if constexpr(!std::is_same<UserType, Boolean>::value) {
    b = UserType(a);
  }
  else {
    // special case for Boolean
    b = bool(a);
  }
}
```
Since code like this is relatively rare (especially in application code
as opposed to lower level framework/library code), this downside is
accepted in favour of the improved usability in the majority of the use
cases.

### DIFF
--- a/device/include/ScalarRegisterAccessor.h
+++ b/device/include/ScalarRegisterAccessor.h
@@ -23,8 +23,11 @@ namespace ChimeraTK {
    *  Note: Transfers between the device and the internal buffer need to be
    * triggered using the read() and write() functions before reading from resp.
    * after writing to the buffer using the operators.
+   *
+   * The second template argument TAG is only used to implement the template specialisation for ChimeraTK::Boolean
+   * without duplicating all code.
    */
-  template<typename UserType>
+  template<typename UserType, typename TAG = std::nullptr_t>
   class ScalarRegisterAccessor : public NDRegisterAccessorAbstractor<UserType> {
    public:
     /** Constructor. @attention Do not normally use directly.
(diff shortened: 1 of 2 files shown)


mhier:
Note, when this PR gets accepted, the ApplicationCore-MicroDAQ package will require a minor code change (replace explicit conversion to `ChimeraTK::Boolean` with conversion `bool`).

killenb:
Wouldn't it be easier to add the implicit conversion operator to all ScalarRegisterAccessor implementations?
We would need a special implementation for string, but this has a template specialisation anyway.

mhier:
> Wouldn't it be easier to add the implicit conversion operator to all ScalarRegisterAccessor implementations? We would need a special implementation for string, but this has a template specialisation anyway.

It would break a lot of code, since implicit conversion from e.g. `ScalarRegisterAccessor<int16_t>` to `int32_t` would become ambiguous (could convert either via `int16_t` or via `bool`).

killenb (approved):

phako (approved):

PR #219, opened 2022-02-03
Drothe/wip 9406 redirectbit target

d-rothe:


device_backends/LogicalNameMapping/include/LNMBackendBitAccessor.h
```diff
@@ -63,19 +63,21 @@ namespace ChimeraTK {
...
-      if(it == map.end() || (_accessor = map[RegisterPath(info->registerName)].accessor.lock()) == nullptr) {
-        _accessor = targetDevice->getRegisterAccessor<uint64_t>(
-            RegisterPath(info->registerName), numberOfWords, wordOffsetInRegister, {});
+      if(it == map.end() || (_accessor = map[key].accessor.lock()) == nullptr) {
+        _accessor = targetDevice->getRegisterAccessor<uint64_t>(key.second, numberOfWords, wordOffsetInRegister, {});
         if(_accessor->getNumberOfSamples() != 1) {
           throw ChimeraTK::logic_error("LNMBackendBitAccessors only work with registers of size 1");
         }
-        map[RegisterPath(info->registerName)].accessor = _accessor;
+        map[key].accessor = _accessor;
```
mhier, line 78: I just realise that this access is not protected by a mutex (or am I missing it only now)? The creation of accessors should be thread safe, I think. You did not introduce this problem here, but maybe we should better add a protection along with this change? The mutex needs to be locked before the first access to the map.

mhier (approved):
Your changes are fine, but I think I found another bug. See my comment above :-)

device_backends/LogicalNameMapping/include/LNMBackendBitAccessor.h
d-rothe: Agreed ... see new commit

PR #220, opened 2022-02-10
Keepalive timer rework

mhier:
This was an old branch from Geogin, which I rebased to the latest master. There are some unmerged changes left. I am unsure whether this should be merged and why it never was.


mhier:
closing for now. It seems this was uncompleted work. Need to see first how to proceed.

PR #221, opened 2022-02-18
added Void changes to spec

vaeng:


doc/spec_TransferElement.dox
```diff
@@ -101,8 +99,11 @@ This document describes the behaviour of the TransferElement base class, the NDR
     - float
     - double
     - std::string
+    - ChimeraTK::Boolean
+    - ChimeraTK::Void
   - 2.2 Applications select a data type for the buffer via the UserType template argument of Device::getZzzRegisterAccessor() resp. DeviceBackend::getRegisterAccessor().
   - 2.3 If needed, the register value is converted between the actual register's data type and the application-selected UserType.
+    - 2.3.1 When the actual type is Void, the conversion is 0 or false in the respective UserType.
```
killenb, line 106: Not only when the actual type is ChimeraTK::Void (I would be very explicit here to distinguish from C++ void). It goes both ways.
If the actual type is void (can the actual type be ChimeraTK::Void, or is it conceptually void?), the buffer is 1x1 of the user type and 0 when reading. When written, the data is discarded.
If the actual type is not void, but the UserType is ChimeraTK::Void, the user buffer has the same shape (number of elements and channels) as the underlying data. When writing, 0 is written to all data elements. Reading is only possible with wait_for_new_data (see 5.2.9)

doc/spec_TransferElement.dox
```diff
@@ -385,6 +386,12 @@ This document describes the behaviour of the TransferElement base class, the NDR
     - 5.2.7 A write operation is executed on a transfer element that cannot be written
       - 5.2.7.1 Check with TransferElement::isWriteable() or TransferElement::isReadOnly()
       - \anchor transferElement_C_5_2_7_2 5.2.7.2 Thrown in doPreWrite() [\ref UnifiedTest_TransferElement_C_5_2_7_2 "U"]
+    - 5.2.8 Scalar, 1D or 2D registers of UserType Void
```
killenb, line 389: - Try not to use the enumeration as sub-header. Directly write the testable sentence.
- Be very exact: It's not Scalar register, but ScalarRegisterAccessor

This section is about when a runtime error is thrown. Each point follows "They are thrown if and only if one of the following conditions are met:" and must be phrased like that.

- 5.2.8 ScalarRegisterAccessor, OneDRegisterAccessor of TwoDRegisterAccessor are instantiated with UserType ChimeraTK::Void. *footnote* (Reason: The data accessing members conceptually don't make sense. Some of them like operator++ etc are not even imeplemtable. VoidRegisterAccessor must me used instead)
   - Throws in the constructor

doc/spec_TransferElement.dox
```diff
@@ -385,6 +386,12 @@ This document describes the behaviour of the TransferElement base class, the NDR
...
+      - The instantiation of the Void UserType is limited to the respective VoidRegister
+      - Other register types will throw an error upon instantiaion.
+    - 5.2.9 UserType Void in combination with read only and not having wait_for_new_data
```
killenb, line 392: See commments to 5.2.8

- read() is called on a VoidRegisterAccessor without wait_for_new_data()
   - Throws in VoidRegisterAccessor::read() *footnote* (Throwing as early as possible means: before even calling the TransferElement implementation)

Note: It is not thrown in the constructor. If you want a VoidRegisterAccessor for writing you don't need the "wait_for_new_data" flag.

Are we missing a point that if you create a read-only accessor. Ah I completely got that wrong. Use "read-only" with dash. I was reading "read is only useful ..". Again, write the sentence "when an exception is thrown".

We need two points
- Throws is the constructor if read-only VoidAccessor is created without "wait_for_new_data"
- Throws in read if called on a VoidAccessor without "wait_for_new_data". Wait, this should be write-only in the catalogue. Ah, because we want to throws as early as possible this is already done in the Abstractor.

killenb (changes requested):

doc/spec_TransferElement.dox
d-rothe: Is this really about runtime errors? Since template instantiation with UserType=Void happens at compile time, you could already throw an error then if Scalar/1D/2D RegisterAccessor are instantiated. So in the spec, it would simply read "instantiation of Scalar/1D/2D RegisterAccessor with UserType=Void is forbidden". Not referring to any runtime error.

doc/spec_TransferElement.dox
d-rothe: sorry Martin, your comment is confusing... I guess the last part captures what you wanted to say

d-rothe (changes requested):

doc/spec_TransferElement.dox
killenb: Your are right. This section is explicitly NOT about runtime errors but about LOGIC errors. Sorry for the confusion.

doc/spec_TransferElement.dox
vaeng: I agree with the explicit distinction 

```suggestion
    - 2.3.1 ChimeraTK::Void is offered as a trigger mechanism rather than data transfer. Therefore the conversion fills the user buffer with 0 or false in the respective UserType for reading with wait_for_new_data (see 5.2.9) and discards any data for writing.
```

Is this too long?

doc/spec_TransferElement.dox
vaeng: > Is this really about runtime errors? Since template instantiation with UserType=Void happens at compile time, you could already throw an error then if Scalar/1D/2D RegisterAccessor are instantiated. So in the spec, it would simply read "instantiation of Scalar/1D/2D RegisterAccessor with UserType=Void is forbidden". Not referring to any runtime error.

I like this short version, although I would add a footnote. The Dox file is different than what I expected, got to google the structure of footnootes.

doc/spec_TransferElement.dox
vaeng: I have to re-read that comment and make a fresh commit.

doc/spec_TransferElement.dox
killenb: a) It's a bit long. We try to have the spec as "what is" to keep it short. There's not so much "why is" in it.
b) It's not quite correct. There are two use  cases which are mixed up in the current wording.

1. The UserType is ChimeraTK::void, but the device has data in it. In this 0 is written to the device, and data coming from the device is discarded. The user Buffer is of type vector<vector<ctk::Void>> and contains elements, but not data content (that's why 0 is "created" when writing and data is discarded when reading into that buffer)
2. The UserType is some data type, but the device conceptually has a void register. When "reading" (i.e. an interrupt arrives, it is seen in the notification queue, user code calls post-read), 0 is filled into the user buffer because we have to invent some data as there is none from the device. When writing (e.g. triggering an action) the data from the UserBuffer is ignored.

d-rothe: I would like to keep at least some of the "why". It is always better if a specification is understandable, and not every sentence needs to be a testable condition. So I would keep at least the first sentence, "ChimeraTK::Void is offered as a trigger mechanism rather than data transfer." The second part does mix up the two use cases, but it still captures the correct idea: Conversion to void is achieved by throwing away data, and conversion from void is achieved by filling up the requested shape with zeros. We only need a precise wording. E. g. based on Martins comment,
"If the actual type is void , the user buffer is 1x1 of the UserType. It is 0 after reading. When written, its data is discarded.
If the actual type is not void, but the UserType is ChimeraTK::Void, the user buffer has the same shape (number of elements and channels) as the underlying actual data type. When writing, 0 is written to all data elements. Reading is only possible with wait_for_new_data (see 5.2.9)"

vaeng:
New commit, what do you think? @killenb  @d-rothe

doc/spec_TransferElement.dox
```diff
@@ -103,7 +103,8 @@ This document describes the behaviour of the TransferElement base class, the NDR
     - ChimeraTK::Void
   - 2.2 Applications select a data type for the buffer via the UserType template argument of Device::getZzzRegisterAccessor() resp. DeviceBackend::getRegisterAccessor().
   - 2.3 If needed, the register value is converted between the actual register's data type and the application-selected UserType.
-    - 2.3.1 When the actual type is Void, the conversion is 0 or false in the respective UserType.
+    - 2.3.1 When the UserType is ChimeraTK::Void, reading from the device discards any actual data. The resulting user buffer is of type <vectorctk::Void>.
```
killenb, line 106: What is the shape of the buffer?
What happens when writing?

doc/spec_TransferElement.dox
```diff
@@ -103,7 +103,8 @@ This document describes the behaviour of the TransferElement base class, the NDR
...
+    - 2.3.2 Converting into another UserType from ChimeraTK::Void will fill the user buffer with 0, false or the respective value for that type.
```
killenb, line 107: " into another UserType from ChimeraTK::Void"
That's wrong. The user type can either be ctk::Void or something else, but never both in the same buffer, so you cannot convert into it.
This point should be: On the device the "register" is conceptually void (e.g. an interrupt without data, or a function without argument that triggers an action), but the user requests a UserType different than void.
Again missing: What is the shape of the buffer? writing direction

doc/spec_TransferElement.dox
```diff
@@ -386,10 +387,9 @@ This document describes the behaviour of the TransferElement base class, the NDR
     - 5.2.7 A write operation is executed on a transfer element that cannot be written
       - 5.2.7.1 Check with TransferElement::isWriteable() or TransferElement::isReadOnly()
       - \anchor transferElement_C_5_2_7_2 5.2.7.2 Thrown in doPreWrite() [\ref UnifiedTest_TransferElement_C_5_2_7_2 "U"]
-    - 5.2.8 Scalar, 1D or 2D registers of UserType Void
-      - The instantiation of the Void UserType is limited to the respective VoidRegister
-      - Other register types will throw an error upon instantiaion.
-    - 5.2.9 UserType Void in combination with read only and not having wait_for_new_data
+    - 5.2.8 ScalarRegisterAccessor, OneDRegisterAccessor or TwoDRegisterAccessor Instantiation with UserType ChimeraTK::Void is forbidden
```
killenb, line 390: Again wrong wording. This bullet point is the extension of "A logic error is thrown if .. ". You cannot complete this with "instantiation is forbidden".

doc/spec_TransferElement.dox
```diff
@@ -386,10 +387,9 @@ This document describes the behaviour of the TransferElement base class, the NDR
...
+      - Throws in constructor
+    - 5.2.9 UserType Void in combination with read-only while not having wait_for_new_data
```
killenb, line 392: see comment above

killenb (changes requested):

vaeng:
@d-rothe Wir sind noch nicht zufrieden, mit den Angaben zur Dimensionalität der Buffer Dimension in 106 und 107. Findest du, dass das woanders besser passen könnte?

d-rothe:
@vaeng  Die Angaben zur Dimensionalität finde ich gut so. 
Was ich noch vermisse, ist das Verhalten "discards data". Das finde ich jetzt gar nicht mehr. D.h. für read von actual type non-void nach UserType=Void oder write von UserType=non-void nach actual type=void.

d-rothe (changes requested):
This test does not make sense, at least not with current implementation of ScalarRegisterAccessor constructor. The latter has a static assertion which should prevent compilation of this test. 
Not sure whether it makes sense to write a test at all, considering this.

d-rothe (approved):

PR #222, opened 2022-02-21
Wip/9048 bug lmap and subdevice

killenb:


device_backends/include/UnifiedBackendTest.h
```diff
@@ -889,6 +891,8 @@ namespace ChimeraTK {
     test_C_5_3_3();
     test_NOSPEC_valueAfterConstruction();
     test_NOSPEC_backendNotClosedAfterException();
+    test_NOSPES_rawTransfer();
```
phako, line 894: Typo: SPES -> SPEC

tests/executables_src/testLMapBackendUnified.cpp
```diff
@@ -323,12 +332,17 @@ struct RegSingleWordB : ScalarRegisterDescriptorBase<RegSingleWordB> {
...
+              << std::endl;
+    return false;
+  }
 
   const uint32_t increment = 3;
 
   typedef uint32_t minimumUserType;
   typedef int32_t rawUserType;
-  DummyRegisterAccessor<minimumUserType> acc{exceptionDummy.get(), "", "/BOARD.WORD_FIRMWARE"};
+  DummyRegisterAccessor<minimumUserType> acc{exceptionDummy3.get(), "", "/BOARD.WORD_FIRMWARE"};
```
phako, line 341: In a future change, can we somehow properly name the exception dummies? Numbering them is a bit confusing further down on what the difference between 1,2 and 3 is

device_backends/LogicalNameMapping/include/LNMAccessorPlugin.h
```diff
@@ -275,10 +275,16 @@ namespace ChimeraTK { namespace LNMBackend {
 
     // obtain desired target type from plugin implementation
     auto type = getTargetDataType(typeid(UserType));
+    auto info = _info.lock();
```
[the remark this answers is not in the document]
killenb: Yes, I was wondering the same. Bit there is no explicit test needed here. This is just locking a weak pointer. If that fails the further access to the pointer would result in a nulltpr exception.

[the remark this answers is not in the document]
killenb: Ah, I see what you mean. This is not in a test, it's in the code and the parent should have gone, so you cannot lock the weak pointer. I will check that.

tests/executables_src/testLMapBackendUnified.cpp
killenb: Probably yes. I will leave a comment at the creation point.
Although dummy3 is a
- I just needed the same register in another backend
- I hacked the push-type into this map file because dummy 1 is the re-used mtcadummy, and the mtcadummy does not support push type

device_backends/LogicalNameMapping/include/LNMAccessorPlugin.h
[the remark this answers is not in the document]
killenb: I will not do this now to avoid unnecessary merge conflicts. I guess this is one of the places where Martin H.'s new deep-copy mechanism will prevent the weak pointer and use a clone() function to duplicate this accessor (?), so I am expecting this to be touched anyway.

device_backends/include/UnifiedBackendTest.h
killenb: done

tests/executables_src/testLMapBackendUnified.cpp
killenb: I renamed it to make it a bit clearer. In the lmap it's still called target1, 2, 3, though.

phako (approved):

PR #223, opened 2022-02-22
8574: implement dummy interrupt x y

phako:


device_backends/include/DummyBackendBase.h
```diff
@@ -103,17 +104,51 @@ namespace ChimeraTK {
...
+        // in the constructor of the accessor. But passing down the base-class of the accessor is very weird
+        // due to the sort-of CRTP pattern used in this base class.
+        auto controller = std::stoi(match[1].str());
+        auto interrupt = std::stoi(match[2].str());
+        try {
+          auto& interruptsForController = _registerMap->getListOfInterrupts().at(controller);
+          if(interruptsForController.find(interrupt) == interruptsForController.end())
+            throw std::out_of_range("Invalid interrupt for controller");
+        }
+        catch(std::out_of_range&) {
```
d-rothe, line 129: throwing std::out_of_range and catching it again looks weird. I understand that you do it because std::map::at() can also throw. A comment would be appropriate

device_backends/include/DummyBackendBase.h
```diff
@@ -103,17 +104,51 @@ namespace ChimeraTK {
...
+          throw ChimeraTK::logic_error("Invalid controller and interrupt combination (" + match[0].str() + ", " +
+              match[1].str() + ": " + regPathNameStr);
+        }
+
+        // Delegate the other parameters down to the accessor which will throw accordingly, to satisfy the specification
+        auto d = new DummyInterruptTriggerAccessor<UserType>(
+            shared_from_this(), [this, controller, interrupt]() { return triggerInterrupt(controller, interrupt); },
```
d-rothe, line 136: Actually I consider it a design flaw that DummyBackendBase::triggerInterrupt is pure virtual, but still DummyBackendBase is templated. That makes it difficult to use. But you found a solution ;-) 
Here we are lucky that the caputure of "this" overlaps in lifetime with the shared_from_this(), otherwise we would rather need a capture of shared_ptr(this).

d-rothe (approved):
works with my test code :-)

device_backends/include/DummyBackendBase.h
phako: Yeah. Added a comment regarding that.

PR #226, opened 2022-02-25
Wip/9048 bug get as cooked

killenb:
Implement raw reading and writing incl. getAsCooked() and setAsCooked() in unified backend test.
Sensitive to the issues which triggered bug remine 9048.
Resolves bug 9048.


d-rothe (approved):
it's a lot, so not sure if I would detect all problems. But basically makes sense to me. So now we require get|setAsCooked for all Accessors which support raw, right?

killenb:
Yes, each register accessor has it, and if it's not raw it throws. And if it is raw, it has to work.
QtHardMon is using that: It is getting a raw accessor, and then gets the cooked value from it.

PR #236, opened 2022-03-09
SharedDummyBackend remote interrupt trigger

phako:
Test and implementation

### DIFF
--- a/tests/include/sharedDummyHelpers.h
+++ b/tests/include/sharedDummyHelpers.h
@@ -5,7 +5,7 @@
 
 #include <sys/file.h>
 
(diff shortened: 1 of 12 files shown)


tests/include/sharedDummyHelpers.h
```diff
@@ -5,7 +5,7 @@
 
 #include <sys/file.h>
 
-enum MirrorRequest_Type { mirrorRequest_From = 1, mirrorRequest_To, mirrorRequest_Stop };
+enum MirrorRequestType { from = 1, to, stop };
```
d-rothe, line 8: If you change names here, it would be better also to move to scoped enum:
enum class MirrorRequestType {...}; 
Only then, enum names from, to, stop are removed from global scope.

d-rothe (approved):

PR #241, opened 2022-03-24
Wip/double buffering plugin with tests

toseko:
Please refer to ticket:
https://redmine.msktools.desy.de/issues/7983


device_backends/DummyBackend/src/DummyBackend.cc
```diff
@@ -49,6 +49,7 @@ namespace ChimeraTK {
     TRY_REGISTER_ACCESS(_barContents[bar].at(address / sizeof(int32_t)) = data;);
   }
 
+/*
```
mhier, line 52: Remove commented code which has been replaced with new code.

device_backends/DummyBackend/src/DummyBackend.cc
```diff
@@ -102,10 +119,25 @@ namespace ChimeraTK {
...
+  void DummyBackend::setReadCallbackFunction(
+      AddressRange addressRange, boost::function<void(void)> const& readCallbackFunction) {
+    _readCallbackFunctions.insert(
+        std::pair<AddressRange, boost::function<void(void)>>(addressRange, readCallbackFunction));
+  }
+
+  void DummyBackend::runReadCallbackFunctionsForAddressRange(AddressRange addressRange) {
+    std::list<boost::function<void(void)>> callbackFunctionsForThisRange =
+        findReadCallbackFunctionsForAddressRange(addressRange);
+    for(std::list<boost::function<void(void)>>::iterator functionIter = callbackFunctionsForThisRange.begin();
```
mhier, line 131: Improve readability / modernise C++:
```
for(auto &function : callbackFunctionsForThisRange) {
  function();
}
```

device_backends/DummyBackend/src/DummyBackend.cc
```diff
@@ -144,6 +176,33 @@ namespace ChimeraTK {
     return returnList;
   }
 
+  std::list<boost::function<void(void)>> DummyBackend::findReadCallbackFunctionsForAddressRange(AddressRange addressRange) {
+    // as callback functions are not sortable, we want to loop the multimap only
+    // once.
+    // FIXME: If the same function is registered more than one, it may be executed
```
mhier, line 182: This can be fixed by changing the `std::list` into an `std::set`

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
```diff
@@ -0,0 +1,124 @@
+#include "LNMDoubleBufferPlugin.h"
+
+#include "BackendFactory.h"
+
+std::string getValue(const std::map<std::string, std::string>& m, std::string key) {
```
mhier, line 5: place in anonymous namespace to avoid conflict with a `getValue()` function declared elsewhere.

mhier: alternatively, just use `std::map::at()` instead, and catch and rethrow the exception (only one catch+rethrow required for all calls)

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
```diff
@@ -0,0 +1,124 @@
...
+        const std::map<std::string, std::string>& parameters, std::string targetDeviceName) {
+      return boost::make_shared<DoubleBufferAccessorDecorator<UserType>>(backend, target, parameters, targetDeviceName);
+    }
+  };
+  /*************************************************/
+  template<typename UserType, typename TargetType>
+  boost::shared_ptr<NDRegisterAccessor<UserType>> DoubleBufferPlugin::decorateAccessor(
+      boost::shared_ptr<LogicalNameMappingBackend>& backend,
+      boost::shared_ptr<NDRegisterAccessor<TargetType>>& target) const {
+    return DoubleBufferPlugin_Helper<UserType, TargetType>::decorateAccessor(
```
mhier, line 55: This can be simplified with modern C++. No `DoubleBufferPlugin_Helper` helper is required. Just use
```
if constexpr(std::is_same<UserType,TargetType>::value) {
  return boost::make_shared<DoubleBufferAccessorDecorator<UserType>>(backend, target, parameters, targetDeviceName);
}
else {
  assert(false);
}
```

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
```diff
@@ -0,0 +1,124 @@
...
+        getValue(parameters, "enableDoubleBuffering"), 0, 0, {});
+    _currentBufferNumberReg = backend->_devices[targetDeviceName]->getRegisterAccessor<uint32_t>(
+        getValue(parameters, "currentBufferNumber"), 0, 0, {});
+    _secondBufferReg = backend->_devices[targetDeviceName]->getRegisterAccessor<UserType>(
+        getValue(parameters, "secondBuffer"), 0, 0, {});
+  }
+
+  template<typename UserType>
+  void DoubleBufferAccessorDecorator<UserType>::doPreRead(TransferType type) {
+    //accquire a lock in firmware (dissable double buffering)
```
mhier, line 74:
```
// acquire a lock in firmware (disable buffer swapping)
```

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
```diff
@@ -0,0 +1,124 @@
...
+  }
+
+  template<typename UserType>
+  void DoubleBufferAccessorDecorator<UserType>::doPostRead(TransferType type, bool hasNewData) {
+    if(_currentBuffer)
+      _target->postRead(type, hasNewData);
+    else
+      _secondBufferReg->postRead(type, hasNewData);
+
+    //release a lock in firmware (enable double buffering)
```
mhier, line 102:
```
// release a lock in firmware (enable buffer swapping)
```

tests/doubleBuffer.xlmap
```diff
@@ -0,0 +1,36 @@
...
+  </redirectedRegister>
+
+
+
+
+
+
+
+
+<!-- This part is not used in double buffering tests. Seems to be left over from preparing a file based on some template -->
```
mhier, line 20: then please remove it...

tests/executables_src/testDoubleBufferng.cpp
```diff
@@ -0,0 +1,183 @@
...
+using namespace ChimeraTK;
+
+BOOST_AUTO_TEST_SUITE(DoubleBufferingBackendUnifiedTestSuite)
+
+/**********************************************************************************************************************/
+
+static std::string rawDeviceCdd("(ExceptionDummy?map=doubleBuffer.map)");
+static auto backdoor =
+    boost::dynamic_pointer_cast<ExceptionDummy>(BackendFactory::getInstance().createBackend(rawDeviceCdd));
+//static boost::shared_ptr<ExceptionDummy> exceptionDummy;
```
mhier, line 22: remove unused commented code

tests/executables_src/testDoubleBufferng.cpp
```diff
@@ -0,0 +1,183 @@
...
+    currentBuffer->write();
+
+    if(deviceWasOpened) {
+      backdoor->close();
+    }
+  }
+
+  void setForceRuntimeError(bool /*enable*/, size_t) {
+    //    target->throwExceptionRead = enable;
+    //    target->throwExceptionWrite = enable;
```
mhier, line 142: dto.

tests/executables_src/testDoubleBufferng.cpp
```diff
@@ -0,0 +1,183 @@
...
+  }
+};
+
+/*********************************************************************************************************************/
+
+struct MyArea1 {
+  std::string path() { return "/doubleBuffer"; }
+  size_t nElementsPerChannel() { return 10; }
+  size_t address() { return 20; }
+  //uint32_t toRaw(float v) { return v * 65536.F; }
```
mhier, line 153: dto.

tests/executables_src/testDoubleBufferng.cpp
```diff
@@ -0,0 +1,183 @@
...
+/*********************************************************************************************************************/
+
+template<typename Register>
+uint32_t AreaType<Register>::_currentBufferNumber = 0;
+
+BOOST_AUTO_TEST_CASE(testUnified) {
+  //"(logicalNameMap?map=doubleBuffer.xlmap&target=(ExceptionDummy?map=doubleBuffer.map))";
+  std::string lmap = "(logicalNameMap?map=doubleBuffer.xlmap&target=" + rawDeviceCdd + ")";
+
+  //exceptionDummy =
```
mhier, line 175: dto.

mhier (changes requested):

device_backends/DummyBackend/include/DummyRegisterAccessor.h
```diff
@@ -173,6 +173,12 @@ namespace ChimeraTK {
           writeCallback);
     }
 
+    /// Set callback function which is called when the register is readed to (through the normal Device interface)
+    //void setReadCallback(const std::function<void()>& readCallback) {
```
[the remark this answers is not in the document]
toseko: I think this part of code was prepared by Jan Timm in order to prepare some tests, but it was never done. The code is not used, I was not sure whether its needed so for time being I just commentd it, but I think it may be removed.

device_backends/DummyBackend/src/DummyBackend.cc
toseko: This is Jan Timm who commented this part of code. I was not looking into DummyBackend.h/cc files. I will remove it, especially that it looks that commented lines are exact copy of code below (from lines 64-79)

device_backends/DummyBackend/src/DummyBackend.cc
toseko: OK. I changed 
`void DummyBackend::runReadCallbackFunctionsForAddressRange`
and 
`void DummyBackend::runWriteCallbackFunctionsForAddressRange`

device_backends/DummyBackend/src/DummyBackend.cc
toseko: Are we sure that we want to change this behaviour ? I do not fully understand what this code is doing and what it suppose to do. If the same function is registered with two different address ranges, then maybe we want to call it twice. This is change of behavior so I would prefer to understand it better before just implementing change according to FIXME comment from Jan Timm.

mhier: Well, this code is introduced by this pull request. It cannot be used anywhere else yet.

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
toseko: Code refactored. getValue is not needed anymore.  std::map::at() used to search value by key in map.

device_backends/DummyBackend/src/DummyBackend.cc
toseko: Yes, you are right. It seems to be kind of logic copy of `findCallbackFunctionsForAddressRange`, but for read operation (not write). It looks that Jan Timm was preparing DummyBackend for something, but it was probably never finished.  Its also why the method setReadCallback was added to DummyRegisterAccessor . 
Code which uses the readCallback part is used only in testDummyBackend which tests the new codeitself, but it seems not to be used in other places in device access. Question is whether this functionality is needed and we should keep it ?

Generic comment is also that the same FIXME comments is in `findCallbackFunctionsForAddressRange` which already is in the master.

There is also now a little naming incosistency. The callbacks for write are stored in std::multimap _writeCallbackFunctions, but functions which operates on this object are just findCallbackFunctionsForAddressRange. Maybe it should be named then: findWriteCallbackFunctionsForAddressRange.

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
toseko: Very nice hint. 
I redefined the 
`template<typename UserType, typename TargetType>
  boost::shared_ptr<NDRegisterAccessor<UserType>> DoubleBufferPlugin::decorateAccessor(
      boost::shared_ptr<LogicalNameMappingBackend>& backend,
      boost::shared_ptr<NDRegisterAccessor<TargetType>>& target) const  {....}`
according to your idea. `DoubleBufferPlugin_Helper ` is not needed.

Just a side note. The same plugin mechanism is used for other plugins - `ForcePollingReadPlugin_Helper` , `ForceReadOnlyPlugin_Helper` etc. Is is a good oportunity to change it also there or should it be done separatelly ?

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
toseko: OK.

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
toseko: OK.

device_backends/DummyBackend/src/DummyBackend.cc
toseko: Read callback functionality has been removed from this branch. Feature is now in separated branch ([wip/9740-ExtendDummyBackendToSupportReadCallbacks](https://github.com/ChimeraTK/DeviceAccess/tree/wip/9740-ExtendDummyBackendToSupportReadCallbacks)) so I think we can resolve this part of the pull request.

mhier (approved):

device_backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
mhier: I created a ticket: https://redmine.msktools.desy.de/issues/9741

PR #242, opened 2022-03-25
fix lmap catalogue not updating from target info

mhier:
The lmap catalogue was no longer being updated with information from its target backends. This caused the observed issues with the math plugin being forced into the wrong direction.

### DIFF
--- a/device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
+++ b/device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
@@ -206,12 +206,15 @@ namespace ChimeraTK {
 
       std::string devName = lnmInfo.deviceName;
 
-      RegisterInfo target_info(lnmInfo.clone()); //Start with a clone of this info as there is not default constructor
-      // In case the devide is not "this" replace it with the real target register info
+      RegisterInfo target_info(lnmInfo.clone()); // Start with a clone of this info as there is not default constructor
+      // In case the device is not "this" replace it with the real target register info
       if(devName != "this") {
         auto cat = _devices.at(devName)->getRegisterCatalogue();
         if(!cat.hasRegister(lnmInfo.registerName)) continue;
         target_info = cat.getRegister(lnmInfo.registerName);
+      }
+      else {
+        target_info = _catalogue_mutable.getRegister(lnmInfo.registerName);
       }
 
       lnmInfo.supportedFlags = target_info.getSupportedAccessModes();
@@ -233,6 +236,8 @@ namespace ChimeraTK {
         lnmInfo.nChannels = target_info.getNumberOfChannels();
       }
       if(lnmInfo.length == 0) lnmInfo.length = target_info.getNumberOfElements();
+
+      _catalogue_mutable.modifyRegister(lnmInfo);
     }
 
     // update catalogue info by plugins


device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
```diff
@@ -206,13 +206,16 @@ namespace ChimeraTK {
...
-      RegisterInfo target_info(lnmInfo.clone()); //Start with a clone of this info as there is not default constructor
-      // In case the devide is not "this" replace it with the real target register info
+      RegisterInfo target_info(lnmInfo.clone()); // Start with a clone of this info as there is not default constructor
+      // In case the device is not "this" replace it with the real target register info
       if(devName != "this") {
         auto cat = _devices.at(devName)->getRegisterCatalogue();
         if(!cat.hasRegister(lnmInfo.registerName)) continue;
         target_info = cat.getRegister(lnmInfo.registerName);
       }
+      else {
```
killenb, line 216: Isn't this redundant? The target_info can been cloned in line 209 from the exact info you are assigning here.

device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
```diff
@@ -233,6 +236,8 @@ namespace ChimeraTK {
         lnmInfo.nChannels = target_info.getNumberOfChannels();
       }
       if(lnmInfo.length == 0) lnmInfo.length = target_info.getNumberOfElements();
+
```
killenb, line 239: Yes, this part was actually missing. Only the copy was being modified. Seem the tests did not catch that this was missing.

killenb (changes requested):
Hm, I did the review last week, but somehow only the comments showed up.
Maybe it is because I finished it with "comment". I wanted to hear a reply whether I got that redundancy thing right in line 217.

I try "request changes" this time. Maybe then the review is accepted.

device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
mhier: In 209, the logical name mapped info is cloned, not the target info.

killenb: yes, traget_info is never cloned. My wording probably is easy to misunderstand (target info has been cloned .... from).

`lnmInfo` is an item of `_catalogue_mutable` ( we are iterating on it), which is cloned **into** `target_info` in line 209. So the target info is a clone of `lnmInfo`

Line 217 is calling `_catalogue_mutable.getRegisterInfo(lnmInfo.registerName)`, which creates another clone of the info for that register and assigns it to `target_info`. The only difference is that the cloning is done inside the catalogue from the "original" entry, and not from the already existing clone `lnmInfo`. But the content should be identical, thus I think it's redundant.

mhier: I understood your wording. Still, line 217 obtains the register info of the target register of the register being described by `lnmInfo`. The member names in the register info struct are obviously confusing. `lnmInfo.registerName` is not the name of the register but the name of the target register.

killenb (approved):

PR #245, opened 2022-03-31
Implement new multiplexed syntax

phako:
First part of #9441 - Parse the new syntax and provide the same accessors as with the old syntax


tests/executables_src/testMapFileParser.cpp
```diff
@@ -341,4 +341,13 @@ BOOST_AUTO_TEST_CASE(testMapFileWithCommentsParse) {
 
 /*******************************************************************************************************************/
 
+BOOST_AUTO_TEST_CASE(testMapFileNewStyleMuxed) {
+  ChimeraTK::MapFileParser map_file_parser;
+  auto [regcat, mdcat] = map_file_parser.parse("newMuxedDataAcessor.mapp");
```
mhier, line 346: This mapp file seems to be used exclusively in this particular (quite minimalistic) test case. Why can't you use newSequences.map instead?

tests/newSequences.mapp
```diff
@@ -0,0 +1,61 @@
+# name                    number of elements       address          size           bar    width   fracbits    signed
+
+# The sequences with default fixed point interpretation 32 bit int (also the shorter ones)
+TEST.MEM_MULTIPLEXED_INT 15 0 60 0
+TEST.MEM_MULTIPLEXED_INT.0 1 0 4 0
+TEST.MEM_MULTIPLEXED_INT.1 1 4 4 0
+TEST.MEM_MULTIPLEXED_INT.2 1 8 4 0
+
+# There are only 15 entries in the sequence, but the AREA has to be a multiple
+# of 4 bytes (32 bits) in order to work with the librabry and the drivers.
```
mhier, line 10: This should not be the case any more. We should be able to do "strange" register sizes and the backend is supposed to do the closest possible transfer size to include all requested bytes.

tests/newSequences.mapp
```diff
@@ -0,0 +1,61 @@
+# name                    number of elements       address          size           bar    width   fracbits    signed
```
mhier, line 1: It seems the named sequences are only appearing in the minimalistic test case which merely counts the number of registers after parsing. I would suggest to change the channel names in this map file to something arbitrary. That change should not make any difference right now.

Also, I would change the order of channel entries in at least one register, such that the lines/entries are not ordered with ascending addresses. This again should not change the test result (since the channels shall be numbered with ascending addresses, not following the line ordering in the map file).

mhier (changes requested):

killenb (approved):

device_backends/NumericAddressedBackend/include/MapFileParser.h
```diff
@@ -95,7 +102,9 @@ namespace ChimeraTK {
     /**
      * Generate 2D register name from main entry for multiplexed registers
      */
-    RegisterPath make2DName(const RegisterPath& pathName);
+    RegisterPath make2DName(const RegisterPath& pathName, const std::string& prefix);
+
+    void createMuxedAccessors(const ParsedLine &pl, std::list<ParsedLine>& channelLines, const std::string& prefix);
```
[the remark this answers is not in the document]
phako: 2 RegisterInfo Objects but yes

tests/newSequences.mapp
phako: Changed in new map file

tests/executables_src/testMapFileParser.cpp
phako: done and extended the test

tests/bam_fmc25_r1225_new_muxed.mapp
```diff
@@ -0,0 +1,101 @@
+APP0.WORD_FIRMWARE_APP                                      1            0            4      1     32      0    0
```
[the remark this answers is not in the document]
phako: Removed

device_backends/NumericAddressedBackend/src/MapFileParser.cpp
```diff
@@ -305,46 +320,56 @@ namespace ChimeraTK {
...
-      bytesPerBlock += plch.nBytes;
-      if(plch.nBytes != 1 && plch.nBytes != 2 && plch.nBytes != 4) {
+    std::vector<NumericAddressedRegisterInfo::ChannelInfo> channels;
+    size_t bytesPerBlock = 0;
+
+    for(auto& channel : channelLines) {
+      channels.emplace_back(NumericAddressedRegisterInfo::ChannelInfo{uint32_t(channel.address - pl.address) * 8,
+          channel.type, channel.width, channel.nFractionalBits, channel.signedFlag});
+      bytesPerBlock += channel.nBytes;
+      if (channel.nBytes != 1 && channel.nBytes != 2 && channel.nBytes != 4) {
```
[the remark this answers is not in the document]
phako: Yes makes sense. Issue 9653

device_backends/NumericAddressedBackend/include/MapFileParser.h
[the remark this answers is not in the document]
phako: Renamed

tests/newSequences.mapp
mhier: Sorry, I didn't realise this. Can you quickly try without the rounding to multiple of 4? It should work as well. If it still doesn't, create a ticket and change the comment to point to that ticket.

mhier (approved):

PR #247, opened 2022-04-04
Wip/9621 math only write with all params

killenb:
The math plugin is only writing if all parameters and the accessor with the math plugin itself have been written once after the device has been opened. This prevents writing several invalid values because the formula has been used without having all data.

### DIFF
--- a/device_backends/LogicalNameMapping/include/LogicalNameMappingBackend.h
+++ b/device_backends/LogicalNameMapping/include/LogicalNameMappingBackend.h
@@ -122,12 +122,10 @@ namespace ChimeraTK {
    private:
     // A version number created when opening the backend. All variables will report this version number until they are changed for the first time
     // after opening the device.
-    ChimeraTK::VersionNumber _versionOnOpen;
-    // We cannot use atomic<VersionNumber> because then share_from_this is not working any more (something with not trivially copyable if I remember correctly)
-    std::mutex _versionOnOpenMutex;
+    std::atomic<ChimeraTK::VersionNumber> _versionOnOpen{ChimeraTK::VersionNumber{nullptr}};
 
    public:
-    ChimeraTK::VersionNumber getVersionOnOpen();
+    ChimeraTK::VersionNumber getVersionOnOpen() const;
   };
 
 } // namespace ChimeraTK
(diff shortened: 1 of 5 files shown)


device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -219,6 +230,20 @@ namespace ChimeraTK { namespace LNMBackend {
 
   /********************************************************************************************************************/
 
+  bool MathPlugin::checkAllParametersWritten(
+      std::map<std::string, boost::shared_ptr<NDRegisterAccessor<double>>> const& accessorsMap) {
+    bool allUpdated = true;
+    auto backend = _backend.lock(); // get a shared pointer from the weak pointer
```
mhier, line 236: Please add a comment why no check is necessary that the lock is successful (you may drop the comment what this line does, as it is very obvious).

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -307,6 +338,9 @@ namespace ChimeraTK { namespace LNMBackend {
       throw ChimeraTK::logic_error("This register with MathPlugin enabled is not writeable: " + _target->getName());
     }
 
+    // The readLatest might throw an exception. In this case preWrite() is never delegated and we must not call the target's portWrite().
+    _skipWriteTransfer = true;
```
mhier, line 342: Maybe rename `_skipWriteTransfer` into `_skipWriteOperation` or `_skipWriteDelegation`? It does not only cause the transfer to be skipped but the entire write operation is not delegated to the `_target`.

mhier (changes requested):

device_backends/LogicalNameMapping/include/LogicalNameMappingBackend.h
```diff
@@ -118,6 +118,16 @@ namespace ChimeraTK {
 
     /** Obtain list of all target devices referenced in the catalogue */
     std::unordered_set<std::string> getTargetDevices() const;
+
+   private:
+    // A version number created when opening the backend. All variables will report this version number until they are changed for the first time
+    // after opening the device.
+    ChimeraTK::VersionNumber _versionOnOpen;
+    // We cannot use atomic<VersionNumber> because then share_from_this is not working any more (something with not trivially copyable if I remember correctly)
```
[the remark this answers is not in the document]
killenb: I am using std::atomic<VersionNumber> now and got rid of the lock

killenb:
I also added the patch with boost::barrier scope to avoid yet another pull request plus review (although it's a feature creep for this pull request)

mhier (approved):

PR #248, opened 2022-04-05
make VersionNumber trivially copyable

mhier:
Does not change any existing behaviour, since the copy constructor was actually already trivial. Now, VersionNumber can be e.g. used in an std::atomic.

See my comments in #247 (may still be pending...)


killenb (approved):

killenb:
I merged this myself so I can proceed with merge request #247

PR #249, opened 2022-04-05
fix ReadAnyGroup::readAnyNonBlocking handling DiscardValueExceptions …

mhier:
…incorrectly

The preRead/postRead calls were not properly matched before, resulting in subtle bugs (deadlocks) when combining with BidirectionalProcessArrays in ApplicatioCore's testable mode.

Note: requires ChimeraTK/cppext#7

### DIFF
--- a/device/include/ReadAnyGroup.h
+++ b/device/include/ReadAnyGroup.h
@@ -496,6 +496,7 @@ namespace ChimeraTK {
   /********************************************************************************************************************/
 
   inline ReadAnyGroup::Notification ReadAnyGroup::waitAnyNonBlocking() {
+  restart_after_discard_value:
     // check if update is available
     if(notification_queue.empty()) {
       // If no notification is present, do not even execute preRead. This is necessary for two reasons:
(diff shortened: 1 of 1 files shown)


killenb (changes requested):
Please increase the required cpp-ext version in CMakeLists.txt

mhier:
> Please increase the required cpp-ext version in CMakeLists.txt

done.

device/include/ReadAnyGroup.h
```diff
@@ -509,6 +510,28 @@ namespace ChimeraTK {
...
+    auto id = notification_queue.front();
+    try {
+      // call to empty() necessary before call to front(), to gain ownership of front element
+      if(push_elements[id].getHighLevelImplElement()->_readQueue.empty()) {
+        // cannot place the call to empty() into the assert, since it must be executed also in release builds
+        assert(false);
+      }
+      push_elements[id].getHighLevelImplElement()->_readQueue.front();
+    }
+    catch(detail::DiscardValueException&) {
```
phako, line 523: Is the behavior correct in a release build? Aren't we going to waitAny without an update or at least some weird inconsistent state in that case?

phako (approved):

PR #258, opened 2022-05-20
support for ASCII registers in numeric addressed backends

mhier:


killenb (approved):
The test only checks basic behaviour with a single string. It could/should be extended to also check the following scenarios:

* Use multiple strings 
* Too long strings are truncated when writing to the device.
* Shorter strings do not leave artefacts. Rest of the buffer is null.
* When writing the max number of bytes no null byte is appended (no buffer overflow)

The last two items can only be tested if a remote value type of vector<char> is used. Currently the test ends when the first null-byte is read (c-string ends).

The implementation does everything correctly, as far as I can see.

mhier:
I added a ticket for the missing tests: https://redmine.msktools.desy.de/issues/9859

PR #261, opened 2022-05-31
Pr/fix interrupt thread safety

killenb:
Use a const reference of the interrupt dispatcher map in NumericAddressedBackend to make sure that acces is thread safe.

Also includes a fix for the version number reported by the device info of PCIe backend. Was broken with version 0.10 if the dummy driver (only worked for minor version between 1 and 9).

### DIFF
--- a/device_backends/pcie/src/PcieBackend.cc
+++ b/device_backends/pcie/src/PcieBackend.cc
@@ -286,7 +286,8 @@ namespace ChimeraTK {
     if(ioctl(_deviceID, _ioctlDriverVersion, &ioctlData) < 0) {
       throw ChimeraTK::runtime_error(createErrorStringWithErrnoText("Cannot read device info: "));
     }
-    os << " DRV VER: " << (float)(ioctlData.offset / 10.0) + (float)ioctlData.data;
+    // major and minor version are in data and offset, respectively
+    os << " DRV VER: " << ioctlData.data << "." << ioctlData.offset;
     return os.str();
   }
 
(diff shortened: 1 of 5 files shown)


device_backends/pcie/src/PcieBackend.cc
```diff
@@ -286,7 +286,8 @@ namespace ChimeraTK {
     if(ioctl(_deviceID, _ioctlDriverVersion, &ioctlData) < 0) {
       throw ChimeraTK::runtime_error(createErrorStringWithErrnoText("Cannot read device info: "));
     }
-    os << " DRV VER: " << (float)(ioctlData.offset / 10.0) + (float)ioctlData.data;
+    // major and minor version are in data and offset, respectively
+    os << " DRV VER: " << ioctlData.data << "." << ioctlData.offset;
```
mhier, line 290: This is a unrelated change, isn't it? Please don't squash merge in that case...

tests/executables_src/testPcieBackend.cpp
```diff
@@ -281,7 +281,7 @@ void PcieBackendTest::testReadDeviceInfo() {
 
   std::string deviceInfo;
   deviceInfo = _pcieBackendInstance->readDeviceInfo();
-  BOOST_CHECK(referenceInfo.str() == deviceInfo);
+  BOOST_CHECK_EQUAL(referenceInfo.str(), deviceInfo);
```
mhier, line 284: I would actually propose instead:
```
  BOOST_TEST(referenceInfo.str() == deviceInfo);
```

PR #271, opened 2022-08-01
Mhier/wip 9490 formal code cleaning

mhier:
Most of the changes should not really change behaviour. Only one commit "fix wrong type information for unsigned int" is actually a bug fix.

**Do not squash** when merging, since the individual commits might help debugging in case of regressions.


mhier:
I decided to keep the accessor abstractor constructors from shared_ptr<NDRegisteAccessor> implicit, because this is actually used in a couple of places (ApplicationCore) where we just would have to repeat the type unnecessarily, while I see no real danger of the implicit conversion going wrong.

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
```diff
@@ -88,19 +85,27 @@ namespace ChimeraTK {
...
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
     read(static_cast<uint8_t>(bar), static_cast<uint32_t>(address), data, sizeInBytes);
+#pragma GCC diagnostic pop
   }
 
   /********************************************************************************************************************/
 
   void NumericAddressedBackend::write(uint64_t bar, uint64_t address, int32_t const* data, size_t sizeInBytes) {
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wdeprecated-declarations"
```
killenb, line 98: I think this is the only place where the 32-bit write is being called at all. Not sure whether the backend programmer will see that he/she is overriding a deprecated function. It might be that this disables all the deprecation warnings which have just been introduced.

However, I agree that this needed and correct here. Was just a remark.

killenb (approved):

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
mhier: You are right, the `[[deprecated]]` specifier does not help for overrides. The ignore pragma also does not hurt, since it only affects that single line (until the `#pragma GCC diagnostic pop`).

PR #280, opened 2022-10-07
Minor Improvements for new compiler and API versions

phako:
with newer standard library or gcc this is not defined otherwise


phako:
Basically coming from https://github.com/ChimeraTK/QtHardMon/commit/a443c5fbb0207c518af9ec1d06c955ff99ce9459 where there is a newer compiler and it's simpler to use libxml++5.0 than 2.6. I will add proper support for 5.0 (cmake checking for 5.0, then falling back to 2.6) as part of the cmake clean-up for the yocto scripts

Issue #281, opened 2022-10-21
Build fails on macOS 12.6 due to redefinition of std::hash and std::less


smarsching:
The build on macOS 12.6 fails:

```
[  1%] Building CXX object CMakeFiles/ChimeraTK-DeviceAccess.dir/device/src/DataConsistencyGroup.cc.o
In file included from /Users/termi/Documents/ChimeraTK/DeviceAccess/device/src/DataConsistencyGroup.cc:4:
In file included from /Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/DataConsistencyGroup.h:5:
In file included from /Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/TransferElementAbstractor.h:5:
In file included from /Users/termi/Documents/ChimeraTK/DeviceAccess/device_backends/include/TransferElement.h:8:
/Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/TransferElementID.h:58:24: error: 
      reference to 'hash' is ambiguous
    friend struct std::hash<TransferElementID>;
                       ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/typeindex:89:50: note: 
      candidate found by name lookup is 'std::__1::hash'
template <class _Tp> struct _LIBCPP_TEMPLATE_VIS hash;
                                                 ^
/Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/TransferElementID.h:10:10: note: 
      candidate found by name lookup is 'std::hash'
  struct hash;
         ^
/Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/TransferElementID.h:61:24: error: 
      reference to 'less' is ambiguous
    friend struct std::less<TransferElementID>;
                       ^
/Library/Developer/CommandLineTools/usr/bin/../include/c++/v1/__functional_base:50:29: note: 
      candidate found by name lookup is 'std::__1::less'
struct _LIBCPP_TEMPLATE_VIS less : binary_function<_Tp, _Tp, bool>
                            ^
/Users/termi/Documents/ChimeraTK/DeviceAccess/device/include/TransferElementID.h:13:10: note: 
      candidate found by name lookup is 'std::less'
  struct less;
         ^
In file included from /Users/termi/Documents/ChimeraTK/DeviceAccess/device/src/DataConsistencyGroup.cc:4:
[...]

PR #282, opened 2022-10-21
Fix build on macOS

smarsching:
There were a couple of problems when building on macOS (12.6), as described in #281. This MR fixes these problems.

### DIFF
--- a/device/include/TransferElementID.h
+++ b/device/include/TransferElementID.h
@@ -4,14 +4,7 @@
 
 #include <cstddef>
 #include <iosfwd>
-
-namespace std {
-  template<class Key>
-  struct hash;
-
-  template<class T>
-  struct less;
-} // namespace std
+#include <functional>
 
 namespace ChimeraTK {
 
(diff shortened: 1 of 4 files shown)


phako:
Is that with target Apple silicon or x86 (mostly talking about the size_t stuff there)

smarsching:
It is on x86_64. I currently do not have access to an ARM-based system to test it there.

mhier (approved):

mhier:
@killenb Can we merge this already or would it disturb the focus work?

mhier:
> Is that with target Apple silicon or x86 (mostly talking about the size_t stuff there)

I guess this is a weird C++ thing: "unsigned long long int" is not the same type as "unsigned long int" despite both being unsigned 64 bit integers and hence having the same value range and bit representation. uint64_t probably can be typedef'd either way, depending on the system headers.

killenb (approved):
The changes should not affect any work in the focus. I don't expect any merge conflicts.

PR #283, opened 2022-10-25
Wip/8141 add convenience functions for accessor read/write

vaeng:
The `ScalarRegisterAccessor` was extended with two methods: 

- `void setAndWrite(UserType newValue, VersionNumber versionNumber`
- `UserType readAndGet()`

Test cases have been added for the two above and also for: 
- `void writeIfDifferent(UserType newValue, VersionNumber versionNumber)`

Special `writeCounting` backend for `writeIfDifferent` implemented.

`setAndWrite` was implemented as a standalone method and not integrated into write to avoid confusion between `VersionNumber` and `UserType` data.

### DIFF
--- a/device/include/ScalarRegisterAccessor.h
+++ b/device/include/ScalarRegisterAccessor.h
@@ -104,6 +104,10 @@ namespace ChimeraTK {
      */
     void writeIfDifferent(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
 
+    void setAndWrite(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
+
+    UserType readAndGet();
+
     friend class TransferGroup;
 
     using NDRegisterAccessorAbstractor<UserType>::get;
@@ -237,6 +241,23 @@ namespace ChimeraTK {
   }
 
   /********************************************************************************************************************/
+
+  template<typename UserType, typename TAG>
+  void ScalarRegisterAccessor<UserType, TAG>::setAndWrite(UserType newValue, VersionNumber versionNumber) {
+    operator=(newValue);
+    if(versionNumber == VersionNumber{nullptr}) versionNumber = {};
+    this->write(versionNumber);
+  }
+
+  /********************************************************************************************************************/
+
+  template<typename UserType, typename TAG>
+  UserType ScalarRegisterAccessor<UserType, TAG>::readAndGet() {
+    this->read();
+    return get()->accessData(0, 0);
+  }
+
+  /********************************************************************************************************************/
   /********************************************************************************************************************/
 
   inline ScalarRegisterAccessor<std::string>::ScalarRegisterAccessor(
@@ -269,7 +290,6 @@ namespace ChimeraTK {
   }
 
   /********************************************************************************************************************/
-
   // Do not declare the template for all user types as extern here.
   // This could avoid optimisation of the inline code.
 
--- a/device/include/ScalarRegisterAccessor.h
+++ b/device/include/ScalarRegisterAccessor.h
@@ -104,8 +104,16 @@ namespace ChimeraTK {
      */
     void writeIfDifferent(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
 
+    /**
+     * Convenience function to set and write new value. The given version number
+     * is only used in case the value differs. If versionNumber == {nullptr}, a new version number is generated only if
+     * the write actually takes place.
+     */
     void setAndWrite(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
 
+    /**
+     * Convenience function to read and return a value of UserType.
+     */
     UserType readAndGet();
 
     friend class TransferGroup;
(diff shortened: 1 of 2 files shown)


device/include/ScalarRegisterAccessor.h
```diff
@@ -104,6 +104,18 @@ namespace ChimeraTK {
      */
     void writeIfDifferent(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
 
+    /**
+     * Convenience function to set and write new value. The given version number
+     * is only used in case the value differs. If versionNumber == {nullptr}, a new version number is generated only if
```
killenb, line 109: The comment is wrong (copy and paste error from writeIfDifferent). The value is always written, not only when it's different.
Hence, the whole scheme with nullprt is not needed.

device/include/ScalarRegisterAccessor.h
```diff
@@ -104,6 +104,18 @@ namespace ChimeraTK {
...
+     * the write actually takes place.
+     */
+    void setAndWrite(UserType newValue, VersionNumber versionNumber = VersionNumber{nullptr});
```
killenb, line 112: just have `VersionNumber versionNumber = {}` as second argument. There will always a new version number be needed.

device/include/ScalarRegisterAccessor.h
```diff
@@ -236,6 +248,23 @@ namespace ChimeraTK {
     }
   }
 
+  /********************************************************************************************************************/
+
+  template<typename UserType, typename TAG>
+  void ScalarRegisterAccessor<UserType, TAG>::setAndWrite(UserType newValue, VersionNumber versionNumber) {
+    operator=(newValue);
+    if(versionNumber == VersionNumber{nullptr}) versionNumber = {};
```
killenb, line 256: You can get rid of this line altogether if the default version number is {}.

tests/executables_src/testScalarRegisterAccessor.cpp
```diff
@@ -135,6 +162,15 @@ BOOST_AUTO_TEST_CASE(testIntRegisterAccessor) {
   accessor.write();
   BOOST_CHECK(dummy == 119);
 
+  // test readAndGet
+  accessor = 470;
+  accessor.write();
+  BOOST_CHECK(accessor.readAndGet() == 470);
+
+  // test setAndWrite
+  accessor.setAndWrite(4711);
```
killenb, line 171: This test is not testing setAndWrite with a version number.  The version number has to be checked for both cases (default argument and given).

tests/executables_src/testScalarRegisterAccessor.cpp
```diff
@@ -288,3 +324,48 @@ BOOST_AUTO_TEST_CASE(testUniqueID) {
...
+  size_t counterBefore = backend->writeCount;
+  accessor.writeIfDifferent(501); // should not write
+  size_t counterAfter = backend->writeCount;
+  BOOST_CHECK(counterBefore == counterAfter);
+
+  // writeIfDifferent with different Value
+  counterBefore = backend->writeCount;
+  accessor.writeIfDifferent(502); // should write
+  counterAfter = backend->writeCount;
+  BOOST_CHECK(counterBefore != counterAfter);
```
killenb, line 359: Probably ok. Maybe a more precise check (counterAfter > counterBefore) or even (counterAfter = counterBefore +1) would be better.

tests/executables_src/testScalarRegisterAccessor.cpp
```diff
@@ -288,3 +324,48 @@ BOOST_AUTO_TEST_CASE(testUniqueID) {
...
+
```
killenb, line 360: Tests with giving a particular version number are missing.

tests/executables_src/testScalarRegisterAccessor.cpp
```diff
@@ -288,3 +324,48 @@ BOOST_AUTO_TEST_CASE(testUniqueID) {
...
+  // test writeIfDifferent for newly created accessor:
+  ScalarRegisterAccessor<int> freshAccessor = device.getScalarRegisterAccessor<int>("APP0/WORD_STATUS");
+  counterBefore = backend->writeCount;
+  freshAccessor.writeIfDifferent(0); // should write
```
killenb, line 364: Especially here check that the version number of the freshly created accessor is {nullptr} (should already be tested somewhere else, though), and test that is changes after writing. Also, test with and without explicitly giving a version number.

killenb (changes requested):
The implementation is OK, however it can be simplified.
The test coverage should be improved to test the version number argument. At one point the test even tests nothing.

vaeng:
Thanks for the comments. The tests are now stricter and check for version number changes.

tests/executables_src/testScalarRegisterAccessor.cpp
```diff
@@ -135,6 +162,15 @@ BOOST_AUTO_TEST_CASE(testIntRegisterAccessor) {
   accessor.write();
   BOOST_CHECK(dummy == 119);
 
+  // test readAndGet
+  accessor = 470;
+  accessor.write();
+  BOOST_CHECK(accessor.readAndGet() == 470);
```
killenb, line 168: Hm, somehow a review comment has not been saved.

This line is only testing half of the story: It is only testing that the content of the buffer is returned. But it does not test the reading is happening because the buffer was 470 already before.

The correct implementation would be
```
dummy = 470;
BOOST_CHECK(accessor.readAndGet() == 470);
```

killenb (changes requested):

killenb (approved):

PR #288, opened 2022-11-04
Wip/9439 lmap parse hex int

d-rothe:
passes tests, including extended test for reported bug


mhier (approved):
I wonder whether this should be unified somehow with the type converters in SupportedUserTypes.h

d-rothe:
> I wonder whether this should be unified somehow with the type converters in SupportedUserTypes.h
might make sense. But use case is a bit different, since changes here consider only config parsing. I think it's fine to be more strict in config parsing than in conversions happening during runtime.

PR #292, opened 2022-11-16
add Utilities::demangleCxx()

mhier:


phako:
There is already boost::core::demangle, so in theory we do not need to implement it ourselves.

killenb (approved):
Approved.

Sorry, I thought this was obsolete because now the boost demangle is used. I did not notice that printStackTrace has been extended.

PR #293, opened 2022-11-21
fix lmap type hint with math plugins

mhier:


mhier:
Note: This PR consists of two commits (no squashing intended). One of them is merely containing minor changes unrelated to the actual content of the PR. I recommend reviewing the two commits separately.

device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
```diff
@@ -11,7 +11,7 @@
 namespace ChimeraTK {
 
   LogicalNameMappingBackend::LogicalNameMappingBackend(std::string lmapFileName)
-  : hasParsed(false), _lmapFileName(lmapFileName) {
+  : hasParsed(false), _lmapFileName(std::move(lmapFileName)) {
```
killenb, line 14: Hm,  have seen places where having an explicit move was preventing copy elision. I never know where to put it or not.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -80,9 +80,6 @@ namespace ChimeraTK { namespace LNMBackend {
     // check if this is the first call to openHook: perform tasks which could not be done in the constructor because
     // the backend is not yet available there.
     if(_backend._empty()) {
-      // make sure the register info is up to date
```
killenb, line 83: Do you understand why this was here? What was it's intention, and why is it working without? And what did it do to disturb the wanted behaviour?

killenb (approved):

device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
mhier: This is a change suggested by the linter (see https://clang.llvm.org/extra/clang-tidy/checks/modernize/pass-by-value.html).
According to https://stackoverflow.com/questions/51705967/advantages-of-pass-by-value-and-stdmove-over-pass-by-reference this is more efficient than pass-by-reference, if an rvalue is passed as string.

PR #297, opened 2022-12-09
fix runtime_error escaping readAny before postRead

mhier:
This may cause termination of the process due to an unhandled exception
in ApplicationCore-based servers using readAny with direct device
accessors in the group.


device/include/ReadAnyGroup.h
```diff
@@ -350,6 +350,9 @@ namespace ChimeraTK {
     try {
       _owner->push_elements[index].getHighLevelImplElement()->_readQueue.pop_wait();
     }
+    catch(ChimeraTK::runtime_error&) {
+      _owner->push_elements[index].getHighLevelImplElement()->_activeException = std::current_exception();
+    }
     catch(detail::DiscardValueException&) {
```
killenb, line 356: Also boost::thread_interrupted has to be delayed here

mhier: added, thanks for the hint

killenb (approved):

PR #298, opened 2022-12-12
Backend factory deprecate old register signature

mhier:


mhier:
I recommend to merge this only after the release planned for today...

phako (approved):

PR #305, opened 2023-01-11
Wip/drothe 11081 lmap parser bug

d-rothe:
see https://redmine.msktools.desy.de/issues/11081

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,7 @@ Project(ChimeraTK-DeviceAccess)
 
 set(${PROJECT_NAME}_MAJOR_VERSION 03)
 set(${PROJECT_NAME}_MINOR_VERSION 07)
-set(${PROJECT_NAME}_PATCH_VERSION 00)
+set(${PROJECT_NAME}_PATCH_VERSION 01)
 include(cmake/set_version_numbers.cmake)
 
 # C++ compiler flags needed to compile this project and against this project


device_backends/LogicalNameMapping/src/LogicalNameMapParser.cc
```diff
@@ -104,11 +104,12 @@ namespace ChimeraTK {
       T value;
       try {
         // base=0 means auto detect base, depending on prefix
-        long long longVal = std::stoll(valAsString, nullptr, 0);
+        int64_t longVal = std::stoll(valAsString, nullptr, 0);
         if(std::is_unsigned_v<T> && longVal < 0) {
           throw std::out_of_range(std::string("negative!"));
         }
-        if(longVal > (long long)std::numeric_limits<T>::max()) {
+        // we need to exclude upper range check for uint64 since max val does not fit into longVal
```
killenb, line 111: Yes, this fixes the issue. uint64_t is the largest integer container in standard C++, so there is nothing to check here.

device_backends/LogicalNameMapping/src/LogicalNameMapParser.cc
```diff
@@ -104,11 +104,12 @@ namespace ChimeraTK {
...
+        if(!std::is_same_v<T, uint64_t> && longVal > (int64_t)std::numeric_limits<T>::max()) {
```
killenb, line 112: This is only checking positive overflow. To be complete there could be a check that negative values are not too small if T is signed.

killenb (approved):

device_backends/LogicalNameMapping/src/LogicalNameMapParser.cc
d-rothe: I now added an explicit check for the lower bound

PR #308, opened 2023-02-01
Add wireshark dissector for rebot

phako:


phako:
Some images of how it looks like

image
image
image
image

killenb (approved):

phako:
come to think of it, we can define a common place to put a map file to and have the dissector parse that and use it if available.

PR #314, opened 2023-02-17
Use getwuid/geteuid instead of getlogin()

phako:
getlogin() sometimes seems to have issues to properly retrieve the
username

### DIFF
--- a/util/src/ProcessManagement.cpp
+++ b/util/src/ProcessManagement.cpp
@@ -8,6 +8,10 @@
 #include <signal.h>
 #include <string>
 #include <unistd.h>
+#include <errno.h>
+#include <iostream>
+#include <cstring>
+#include <pwd.h>
 
 bool processExists(unsigned pid) {
   return !kill((pid_t)pid, 0);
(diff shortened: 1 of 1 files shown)


phako:
Also seems to work on macOS

killenb (approved):

util/src/ProcessManagement.cpp
```diff
@@ -18,9 +22,17 @@ unsigned getOwnPID(void) {
...
+  if (pwent == nullptr) {
+    // pwent == nullptr and errno == 0 is a valid scenario - There was no error, but the system also
+    // could not find the corresponding user name.
+    if(savedErr != 0) {
+        std::cout << "Failed to lookup user name, expect issues: " << strerror(savedErr) << std::endl;
+    }
+    return std::string{"**unknown*user*name**"};
   }
-  return std::string(login);
+  return std::string{pwent->pw_name};
```
killenb, line 37: Linter warning: don't repeat return type. Just use curly braces.

phako: Updated

PR #319, opened 2023-02-23
Use CMake to check and set -D_FILE_OFFSET_BITS

phako:


killenb (approved):
The previous C++ version was just setting the flag unconditionally. Wouldn't this have been a quick cmake implementation? This one is really fancy, but I think it's clean to only add the flag when it's needed.

PR #329, opened 2023-03-13
Wip/11221 modernize cmake

d-rothe:


d-rothe:
project-template branch needs to be merged first. this branch shows usage.

mhier:
Please do not squash all commits together. We should have two commits:
- [make DeviceAccess example compile again](https://github.com/ChimeraTK/DeviceAccess/pull/329/commits/61bdc9e5cf991a4e3c5fd1b8de0e7eeae792d9db) - just as it is right now (maybe with improved title "fix: make example compile again")
- All other commits into a single commit "chore: modernsie cmake" or so.

mhier (approved):

CMakeLists.txt
```diff
@@ -14,6 +14,7 @@ include(cmake/set_version_numbers.cmake)
 set(${PROJECT_NAME}_CXX_FLAGS "${${PROJECT_NAME}_CXX_FLAGS} -DFUSION_MAX_MAP_SIZE=30 -DFUSION_MAX_VECTOR_SIZE=30 -D_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR")
```
mhier, line 14: As a side note: We can remove -`DFUSION_MAX_MAP_SIZE` and `-DFUSION_MAX_VECTOR_SIZE` here, they are no longer necessary with C++11 and higher. (Otherwise they would need to be exported as public)

d-rothe: they are actually exported, see line 101

d-rothe: I'm removing them. Just testing that everything compiles downstream

PR #332, opened 2023-03-20
fix: make example compile again

d-rothe:
and move to imported targets


killenb (changes requested):
Please split the changes into two commits. I want to make a DeviceAcces release from the state before the CMake changes (merge from project template) and would cherry-pick the C++ changes.

killenb (approved):

killenb:
Use "rebase and merge" when merging into the master. Do not squash.

PR #333, opened 2023-03-20
remove testVirtualFunctionEmplateCompilerFlags

d-rothe:
since compiler flags were removed and test became obsolote


killenb (approved):
There is a leftover check in VirtualFunctionTemplate.h, where at least the error message is outdated. Maybe we should leave the check in, or change it to > 30?

d-rothe:
> There is a leftover check in VirtualFunctionTemplate.h, where at least the error message is outdated. Maybe we should leave the check in, or change it to > 30?

the check in VirtualFunctionTemplate.h was already removed, maybe you are looking at old code (pre-merge of chore: modernize project-template) ?

killenb:
Indeed, I did a ``git pull``, but was on the wip/11221-modernizeCmake branch which is already gone.

PR #341, opened 2023-04-20
Wip/spec interrupt controller handler

killenb:
Unfortunately the wording is very clumsy and difficult to read. In addition there are two fixmes which need further discussion.


tests/monostableTriggerPlugin.xlmap
mhier, line 1: please don't include these changes in this PR, it should be only about the spec IMO...

doc/spec_InterruptControllerHandler.dox
```diff
@@ -0,0 +1,91 @@
...
+
+### Context ###
+
+The current implementation INTERRUPTc:i (with c being the interrupt controller number and i being the interrupt) is wrong on several levels.
+* The highest level is always associated with a hardware interrupt. If there is one interrupt controller, this is correct. If there is no interrupt controller, it is already wrong. It currently is INTERRUPT0:i, but should be INTERRUPTi. (e.g. with interrupt controller INTERRUPT0:5 is on HW interrupt 0, while without it is on HW interrupt 5, which is inconsistent).
+* Nested interrupts cannot be described at all.
+
+This specification describes the hierarchical interrupt scheme that solves these issues.
+
+## A. Definitions ##
```
mhier, line 20: For me, a section "Definitions" would describe terms used by the spec to describe the behaviour. These definitions seem to describe already the implementation, as the keyword "INTERRUPT" is probably found in the map file. I think, at this point of the specification we need to describe the behaviour not the implementation. So what should be defined here are terms like "interrupt controller" and "interrupt handler" etc. (as they are used here in this document).

Also, I think it makes it complicated by weaving the concept of nested interrupts strongly into the fundamental principles and wording. I would start to introduce the concept of a single interrupt (aka "INTERRUPTi" but ideally without the map file syntax), then based on that introduce the concept of an interrupt controller (aka "[baseInterrupt]:k", again ideally free of the map file syntax). The concept of nested interrupts then follows natively essentially by induction.

My second paragraph here in this comment is actually more directed already towards the description of the behaviour. In this Definitions section we should only concentrate on the definitions which are used below and merely describe what they mean (not how they interact).

doc/spec_InterruptControllerHandler.dox
```diff
@@ -0,0 +1,91 @@
...
+
+- hwInterruptID is an integer which identifies the primary hardware interrupt
+- qualifiedID is the ID including all the parent hierarcy. It can be evaluated on any level.
+  Example: INTERRUPT2:3:4 on a nested interrupt controller
+  - 2 is the qualifiedID of the parent interrupt controller (identical to the HW interrupt ID, see \ref interruptHandler_B1_1).
+  - 2:3 is the qualified ID of the nested interrupt controller (identical to the interrupt generated by interrupt controller 2)
+  - 2:3:4 is the qualified ID of the interrupt generated by controller 2:3
+- INTERRUPTi[:j[:k[...]]] means any level of nesting
+
+- The interrupt controller resides in firmware
```
mhier, line 30: This should go into a comment/footnote as an explanation. It is not part of the spec in the sense that we actually do not care. It could be implemented in software, similar like any register can be a software register (cf. DOOCS backend) - it just would make little sense to use this concept there (unless maybe abstraction is required?).

I think, what you actually mean here: Interrupt controller is on the sending end of the interrupt, while the InterruptControllerHandler class is at the receiving end.

doc/spec_InterruptControllerHandler.dox
```diff
@@ -0,0 +1,91 @@
...
+
+## B. Behavioural specification
+
+- B1. Interrupts and interrupt controllers are identified with a numerical ID (unsigned integer)
+  - \anchor interruptHandler_B1_1 B1.1 The controller ID is always identical to the parent interrput (e.g. controller 1 is behind hardware interrupt 1)
+- B2. To identify interrupts resolved by a controller, their IDs are appended to the parent interrupt in a hierarchical way. This allows to nest interrupt controllers.
+- B3. In string form (e.g. in the map file and this document), a colon ":" is used as hierarchy separator.
+  - B3.1 Programmatic implementations can vary from this (e.g. in C++ the interrupt can be descibed as a list/vector of integers)
+- B4. It is always possible to get an accessor for all interrupt hierarchy levels.
+  - B4.1 An accessor with the name INTERRUPTk[:l[:m[...]]] can be acquired, even if there is no map file entry for it.
```
mhier, line 43: Ah, so now "INTERRUPT..." is a register name? This should be made clear earlier.

doc/spec_InterruptControllerHandler.dox
```diff
@@ -0,0 +1,91 @@
...
+- B5. Backend implementations (NumericAddressedBackend)
```
mhier, line 44: Is this still behaviour, or a new section for the implementation? This depends on the aim of this document, so I am really unsure...

mhier (commented):
Note: I just had a glance at the first half of the document roughly. It is a bit unclear to me what this document should specify, so I would like to understand that first...

tests/monostableTriggerPlugin.xlmap
killenb: oops, this was by mistake.

killenb:
The whole concept needs rework. Probably it's just an implementation with docu in the base class and no formal spec.

PR #343, opened 2023-04-28
fix: lmap math plugin with push parameter shutdown (#11506)

d-rothe:
discussion: see slides from https://redmine.msktools.desy.de/issues/11506


tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+  setDMapFilePath("mathPluginWithPushPars.dmap");
+  {
+    // we enforce that initial values from variable definition in xlmap are never used in MathPugin
+    // - therefore, in test here, we must issue write for all push-parameters
+    // - then, version number check in MathPlugin will notice valid data has been provided after open
+    // - for poll-type accessors, version number check also succeeds since they get recent version numbers anyway
+    // Test needs to call activateAsyncRead but it must be irrelevant whether before or after writes
+
+    ChimeraTK::Device targetDevice;
+    targetDevice.open("HOLD");
```
killenb, line 50: Isn't this version of open() deprecated already?
```suggestion
    ChimeraTK::Device targetDevice("HOLD");
    targetDevice.open();
```

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+
+    auto accTarget = targetDevice.getScalarRegisterAccessor<uint32_t>("HOLD0/WORD_G");
+    auto pollPar = targetDevice.getScalarRegisterAccessor<uint32_t>("HOLD0/POLLPAR");
+    pollPar = 1;
+    pollPar.write();
+
+    ChimeraTK::Device device;
+    device.open("EOD");
```
killenb, line 58:
```suggestion
    ChimeraTK::Device device("EOD");
    device.open();
```

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+
+    ChimeraTK::Device targetDevice;
+    targetDevice.open("HOLD");
+
+    auto accTarget = targetDevice.getScalarRegisterAccessor<uint32_t>("HOLD0/WORD_G");
+    auto pollPar = targetDevice.getScalarRegisterAccessor<uint32_t>("HOLD0/POLLPAR");
+    pollPar = 1;
+    pollPar.write();
+
+    ChimeraTK::Device device;
```
killenb, line 57: Why do you use two devices at once? This is confusing.

killenb: OK, I get it now. One is the "raw device", i.e. without logical name mapping, the other one is the LMAP-Device. Please make this clear in the names and/or add comments.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    auto accMathWrite = device.getScalarRegisterAccessor<double>("DET/GAIN");
+    // we don't have main value (x in formula) yet, since it wasn't yet written.
+    // therefore, we expect to have no value yet for formula output (0 is default from dummy construction)
+    accTarget.readLatest();
+    BOOST_CHECK_EQUAL(int(accTarget), 0);
+
+    // write to main value and check result
+    accMathWrite = 3;
+    accMathWrite.write();
+    accTarget.readLatest();
```
killenb, line 74: Why do you use readLatest()? This accessor does not have wait_for_new_data().

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    BOOST_CHECK_EQUAL(int(accTarget), 100 * (int)pollPar + 10 * (int)pushPar + (int)accMathWrite);
```
killenb, line 75: don't use C-type casts!

killenb: Why do you cast at all? The implicit conversion  should do the job for the multiplication and addition. int(accTarget) is the only needed one and proper C++ already.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+
+    // write to push-parameter and check result
+    // note, it's a new feature that result is completely written when write() returns.
+    pushPar = 4;
+    pushPar.write();
+    accTarget.readLatest();
```
killenb, line 81: actually, readLatest() is even conceptually wrong here. It means "take whatever has arrived in this thread until now, whatever the other threads have done". What you would want if it was an push-type accessor and you wanted to check that it has been pushed is an explicit ``BOOST_CHECK(accTarget.readLatest());``. A simple read() is enough here, since it cannot block.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    BOOST_CHECK_EQUAL(int(accTarget), 100 * (int)pollPar + 10 * (int)pushPar + (int)accMathWrite);
+
+    // check that MathPlugin waits on all initial values.
```
killenb, line 84: Confusing comment. I just wanted to write "you already did that". But it was only tested that writing the main value without parameter does not propagate.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    accTarget = 0;       // reset result in dummy
+    accTarget.write();
+    pollPar.write(); // restore value in case lost during reopen (actually not required with current dummy impl)
+    device.open("EOD");
+    device.activateAsyncRead();
+
+    accMathWrite = 5;
+    accMathWrite.write();
+    // check that MathPlugin did not yet write to the device - it must wait on push-parameter value
+    accTarget.readLatest();
```
killenb, line 96: Well, actually this was already tested. Except for a "fresh" backend instead of a re-opened one.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    // so test that update is retrieved even if it's called last
+    device.close();
+    device.open();
+    accTarget = 0;
+    accTarget.write();
+    pollPar.write();
+    accMathWrite = 7;
+    accMathWrite.write();
+    pushPar = 6;
+    pushPar.write();
```
killenb, line 113: You have to test that nothing has been written here. This is important to comply to the spec. Only "activateAsyncRead" must cause writing in this test.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    device.activateAsyncRead();
+    accTarget.readLatest();
+    BOOST_CHECK_EQUAL(int(accTarget), 100 * (int)pollPar + 10 * (int)pushPar + (int)accMathWrite);
+  }
+  // regression test for bug https://redmine.msktools.desy.de/issues/11506
+  // (math plugin + push-parameter + shm has resource cleanup problem)
+  BOOST_CHECK(DummyForCleanupCheck::cleanupCalled);
```
killenb, line 120: Actually, cleaupCalled is never set to false.

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -0,0 +1,125 @@
...
+    device.open();
+    accTarget = 0;
+    accTarget.write();
+    pollPar.write();
+    accMathWrite = 7;
+    accMathWrite.write();
+    pushPar = 6;
+    pushPar.write();
+    device.activateAsyncRead();
+    accTarget.readLatest();
```
killenb, line 115: What actually is missing is the test that write is only happening once. When reading though the code I thought it would be executed multiple times. It is not, but there is no check here. You could extend the dummy with a write counter on the target register.

killenb (commented):
I still don't understand how this solves the share_pointer problem because the formula helper is holding accessors, which should hold shared pointers to the backend.

It was a bit annoying to get changes pushed while during the review. I had to read through the long changes in LNMMathPlugin.cc three times because it was not clear what has changed.

Do not force push after requesting a review! The review comments will lose the versions they refer to if you change the history of the branch.

I don't give explicit approval because there are too many open questions. But I also don't request changes. Probably only mode discussion/explanation is needed.

device_backends/LogicalNameMapping/include/LNMMathPlugin.h
```diff
@@ -0,0 +1,105 @@
+// SPDX-FileCopyrightText: Deutsches Elektronen-Synchrotron DESY, MSK, ChimeraTK Project <chimeratk-support@desy.de>
+// SPDX-License-Identifier: LGPL-3.0-or-later
+#pragma once
+
+#include "LNMAccessorPlugin.h"
+
+#include <boost/make_shared.hpp>
+
+#include <exprtk.hpp>
```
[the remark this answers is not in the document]
mhier: This is a backend header file which is not part of the public interface. We should probably not even install it (afaik we also do not install any headers in the "external" backends like the DoocsBackend).

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -73,186 +47,143 @@ namespace ChimeraTK::LNMBackend {
...
+    // If write direction, check for push-type parameters if enabled
+    if(_isWrite && _enablePushParameters) {
+      for(const auto& parpair : _parameters) {
+        auto paramFlags = backend->getRegisterCatalogue().getRegister(parpair.second).getSupportedAccessModes();
+        if(paramFlags.has(AccessMode::wait_for_new_data)) {
+          _hasPushParameter = true;
+          break;
         }
       }
     }
```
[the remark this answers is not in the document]
mhier: actually: don't we enforce now that parameters are variables? Don't variables always support wait_for_new_data? Maybe this can simplify the code?

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -361,13 +286,23 @@ namespace ChimeraTK::LNMBackend {
     if(!_p->_isWrite) {
       throw ChimeraTK::logic_error("This register with MathPlugin enabled is not writeable: " + _target->getName());
     }
+    // LNM backend - we know it exists
+    auto backend = h->_backend.lock();
+    if(!backend->isOpen()) {
```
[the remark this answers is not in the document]
mhier: According to the [TransferElement spec Section C.2](https://chimeratk.github.io/DeviceAccess/master/spec__transfer_element.html#transferElement_C_2), we are not allowed to throw any new exception in the post phase. This may cause issues with other decorators (e.g. ExceptionHandlingDecorator in ApplicationCore).

tests/executables_src/testLMapMathPluginPushPars.cc
mhier: You have to cast accessors in BOOST_CHECK_EQUAL etc, otherwise you get some errors because BOOST tries to cast them into a string for the message, which does not work. It is likely sufficient though to do that for the accessor which is not part of any expression (`accTarget` in this case).

I also recommend to use `BOOST_TEST` instead (although it has the same issue with casting):

```
BOOST_TEST(int(accTarget) ==  100 * pollPar + 10 * pushPar + accMathWrite);
```

device_backends/LogicalNameMapping/include/LNMBackendVariableAccessor.h
```diff
@@ -41,6 +47,10 @@ namespace ChimeraTK {
 
     void interrupt() override;
 
+    /// to be called only internally by LNMBackend. This deletes the shared ptr count to the LNM backend,
```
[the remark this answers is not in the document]
d-rothe: Yes this line is outdated and I removed it finally, when cleaning up old stuff.

device_backends/LogicalNameMapping/include/LNMBackendVariableAccessor.h
```diff
@@ -150,6 +163,18 @@ namespace ChimeraTK {
       });
     }
 
+    // make sure FormulaHelpers for MathPlugin instances involving this variable as push-parameter are created
+    auto& lnmVariable = _dev->_variables[_info.name];
+    for(auto* mp : lnmVariable.usingFormulas) {
+      if(mp->_hasPushParameter) {
+        // following check eliminates recursion, getFormulaHelper also creates Variable accessors
```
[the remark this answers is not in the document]
d-rothe: There is not many places where _creatingFormulaHelper is accessed. Also, if you do it wronly and get infinite recursion you will notice immediately, since it causes a crash.

device_backends/LogicalNameMapping/include/LNMBackendVariableAccessor.h
```diff
@@ -226,6 +251,27 @@ namespace ChimeraTK {
 
   /********************************************************************************************************************/
 
+  template<typename UserType>
+  void LNMBackendVariableAccessor<UserType>::doPostWrite(
+      TransferType /*type*/, ChimeraTK::VersionNumber /* versionNumber */) {
+    // call write functions which make use of this parameter in MathPlugin-handled formulas
+
+    auto& lnmVariable = _dev->_variables[_info.name];
```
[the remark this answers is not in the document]
d-rothe: ok, will simplify it.

device_backends/LogicalNameMapping/include/LNMMathPlugin.h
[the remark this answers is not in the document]
d-rothe: I think the whole LogicalNameMapping/include folder should not get installed. Maybe even also not for some other backends?

device_backends/LogicalNameMapping/src/LNMAccessorPlugin.cc
```diff
@@ -3,6 +3,7 @@
 
 #include "LNMBackendRegisterInfo.h"
 #include "LNMDoubleBufferPlugin.h"
+#include "LNMMathPlugin.h"
```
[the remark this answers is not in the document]
d-rothe: not sure why it was working before, it's needed in line 20 for make_shared<MathPlugin>

d-rothe:
> I still don't understand how this solves the share_pointer problem because the formula helper is holding accessors, which should hold shared pointers to the backend.
> 
> It was a bit annoying to get changes pushed while during the review. I had to read through the long changes in LNMMathPlugin.cc three times because it was not clear what has changed.
> 
> Do not force push after requesting a review! The review comments will lose the versions they refer to if you change the history of the branch.
> 
> I don't give explicit approval because there are too many open questions. But I also don't request changes. Probably only mode discussion/explanation is needed.

Sorry for the force push. I was not aware you had already started the review; will avoid it in the future.
The cleanup process can be understood looking at slide 3 of LNM-mathplugin-cleanup.odp, attached at https://redmine.msktools.desy.de/issues/11506 
You are right, FormulaHelper is holding accessors, which again hold the backend. But FormulaHelper is destructed as soon as the last accessor from the outside (user writing to x or formula parameter) is thrown away.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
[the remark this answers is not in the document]
d-rothe: I don't particularly like the idea of allowing only variables as parameters. I have use-cases for redirectedRegister parameters:
dcm_ctrl_server uses overall status calculated from several hw registers (this is read direction, though).
Don't you think it is confusing if we allow different things for read than for write direction?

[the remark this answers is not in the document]
d-rothe: I don't mind throwing if enable_push_parameters is requested but there is none.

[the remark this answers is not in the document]
d-rothe: currently push_parameters are only supported for write direction. I think we should also throw if someone wants them for read direction.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -301,7 +234,7 @@ namespace ChimeraTK::LNMBackend {
   template<typename UserType>
   MathPluginDecorator<UserType>::MathPluginDecorator(boost::shared_ptr<LogicalNameMappingBackend>& backend,
       const boost::shared_ptr<ChimeraTK::NDRegisterAccessor<double>>& target,
-      const std::map<std::string, std::string>& parameters, MathPlugin* p)
+      const std::map<std::string, std::string>& /* parameters */, MathPlugin* p)
```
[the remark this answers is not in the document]
d-rothe: I now simplified it.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -310,20 +243,12 @@ namespace ChimeraTK::LNMBackend {
...
-    // obtain accessors for parameters
-    std::map<std::string, boost::shared_ptr<ChimeraTK::NDRegisterAccessor<double>>> accessorMap;
-    for(const auto& par : parameters) {
-      accessorMap[par.first] = backend->getRegisterAccessor<double>(par.second, 0, 0, {});
-    }
-
-    // compile formula
-    h.varName = this->getName();
-    h.compileFormula(_p->_formula, backend, accessorMap, target->getNumberOfSamples());
+    h = _p->getFormulaHelper(backend);
```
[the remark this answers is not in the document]
d-rothe: renamed

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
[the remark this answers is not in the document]
d-rothe: > If I am not mistaken doPostxxx must never throw logic errors according to spec. We might have missed this because we never thought about close(). Probably the right behaviour (according to current spec) would be to throw in close() if there are still pending post-reads(). This might not even be implementable. I guess we have to extend the spec.

You are just commenting a code line for doPreWrite.   MathPluginDecorator<UserType>::doPostWrite indeed does not throw new exceptions, so I think we have no problem here.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -515,10 +441,66 @@ namespace ChimeraTK::LNMBackend {
...
+              info->name + "', parameter name: '" + parpair.first + "'");
+        }
+        _accessorMap[parpair.first] = acc;
+      }
+      _lastMainValue.resize(length);
+    }
+    else {
+      // obtain poll-type accessors for parameters
+      for(const auto& par : _mp->_parameters) {
+        _accessorMap[par.first] = backend->getRegisterAccessor<double>(par.second, 0, 0, {});
```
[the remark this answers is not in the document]
d-rothe: True point. I guess I was confused because I at some time moved it to MathPlugin. 
eplaced it by shared ptr.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -515,10 +441,66 @@ namespace ChimeraTK::LNMBackend {
...
+    varName = info->name;
+
+    boost::shared_ptr<DeviceBackend> targetDevice;
+    if(info->deviceName == "this") {
+      targetDevice = _backend.lock();
+    }
+    else {
+      targetDevice = BackendFactory::getInstance().createBackend(info->deviceName);
+    }
+    _target = targetDevice->getRegisterAccessor<double>(info->registerName, info->length, info->firstIndex, {});
```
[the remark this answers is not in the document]
d-rothe: Yes

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: there is no warning. But I agree that open(deviceAlias) is bad, since you could then provide a different alias, from last time.
So it fixed it.

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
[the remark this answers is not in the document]
killenb: Hm, I actually meant: Isn't the share pointer loop back if the plugin is holding accessors?

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: Right. Renamed it -> logicalDevice

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: no specific reason. I can change it.

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: changed it to read()

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: I'll update the comment
// re-open and test again, with different write order (x,p) instead of (p,x)

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: here the point was testing with different write order after open

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: ok added the test BOOST_TEST(int(accTarget) == 0);

tests/executables_src/testLMapMathPluginPushPars.cc
killenb: Can you explicitly write ``std::atomic<bool> cleanupCalled{false}`` to make it clear?

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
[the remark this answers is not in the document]
mhier: I though push parameters only work with variables already now?

I agree, enabling push parameters for the reading direction should throw a logic_error.

device_backends/LogicalNameMapping/include/LNMMathPlugin.h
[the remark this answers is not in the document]
d-rothe: I now added device_backends/LogicalNameMapping/include/internal/ which currently has only the problematic class MathPluginFormulaHelper. It's included only from .cc files so no problem for API.
Changed cmake so it does not install this.
This makes sure we have exprtk still private.

device_backends/LogicalNameMapping/src/LogicalNameMappingBackend.cc
```diff
@@ -363,6 +363,25 @@ namespace ChimeraTK {
                 std::terminate();
               }
             }
+            // also give MathPlugin instances which are using the variable as push-parameter the chance to update
+            for(auto* mp : lnmVariable.usingFormulas) {
```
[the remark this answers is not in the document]
d-rothe: Yes, updateResult() is called more often than necessary .. but it shouldn't be a performance issue. If I wanted to improve it, I would need another list which I need to modify from inside the lambda... also ugly.

doc/logicalNameMapper.dox
```diff
@@ -150,14 +150,15 @@ number of elements of the target register. All calculations will be done interna
...
-  is received, all poll-type parameters are polled, the result of the formula is computed (based on the last data
-  written to the logical register) and the target register is written.
+- <code>enable_push_parameters</code>: This parameter takes no value. It enables push-like logic for all paramters
+  which support wait_for_new_data, in particular for variables defined in the logical name map.
+  When writing to these parameters, the formula result is recalculated and immediately written to the target register.
+  A side constraint is that the value <code>x</code> was written as well, since last (re-)opening of the device.
 - Any additional parameter value will be interpreted as a register name of the logical name mapper device. The register
   value will be made available to the formula by the name of the parameter. Everytime the formula is evaluated, the
   registers will be read, so the current values are provided.
+- CAVEAT: initial values for variable parameters are not used! The plugin waits until all parameters as well as the main
```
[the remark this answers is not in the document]
d-rothe: I now removed the 'CAVEAT' word

tests/mathPluginWithPushPars.xlmap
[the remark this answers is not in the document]
d-rothe: Those were coming from EOD server xlmap. But I now renamed them.

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: changed it to use BOOST_TEST

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: done

tests/executables_src/testLMapMathPluginPushPars.cc
d-rothe: Yes, adding this check actually showed a bug. 
But it was not complicated to fix. Only needed to replace list(Math Plugins) -> set(Math Plugins)

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -73,186 +48,151 @@ namespace ChimeraTK::LNMBackend {
...
+        _target->setDataValidity(ChimeraTK::DataValidity::faulty);
+      }
+
+      _target->writeDestructively(versionNumber);
+    }
+    catch(runtime_error&) {
+      // runtime_error from param.readLatest() or target->write()
+      // we could actually even ignore it, since we don't expect exceptions on param.readLatest(), and target->write()
+      // already puts backend into exception state.
+      // TODO discuss - should we direct to lnmBackend->getExceptionBackend() instead?
```
d-rothe, line 165: @mhier  can you comment on this?

d-rothe: I guess _target->getExceptionBackend() will give you LNM backend, but I'm not sure if this in line with the spec. The spec for Exception backend treatment is really hard to get.

mhier: Yes, you should always report an exception to the exception backend. It might be the logical name mapping backend, or it might be a backend on a higher level (e.g. another LNM backend using the LNM backend which handed out the accessor as a target). Higher level backends are required to pass exceptions down to all their target backends. This way we make sure all participating backends are going into an exception state consistently.

d-rothe: Does "yes" mean the code line is fine? More precisely, which is better:
(A)  _target->getExceptionBackend()->setException()
(B) lnmBackend->getExceptionBackend()->setException()

d-rothe: (A) is right, (B) does not make sense because getExceptionBanckend() is for accessors

tests/executables_src/testLMapMathPluginPushPars.cc
```diff
@@ -47,73 +48,120 @@ BOOST_AUTO_TEST_CASE(testPushPars) {
...
+    ChimeraTK::Device logicalDevice("EOD");
+    logicalDevice.open();
+    logicalDevice.activateAsyncRead();
+    auto pushPar = logicalDevice.getScalarRegisterAccessor<uint32_t>("DET/PUSHPAR");
 
     pushPar = 2;
     pushPar.write();
 
-    auto accMathWrite = device.getScalarRegisterAccessor<double>("DET/GAIN");
+    auto accMathWrite = logicalDevice.getScalarRegisterAccessor<double>("DET/X");
```
killenb, line 68: :+1:  X is a very good name for the main variable because it's the same as in the formula

killenb (approved):

PR #347, opened 2023-05-11
fix: map file interrupts syntax

killenb:
The map file syntax was fixed to always have one controller and one level of nested interrupts. This is conceptually wrong.

The new syntax allows primary interrupts without interrupt controllers and nested interrupts with an arbitrary nesting level. The implementation only supports primary interrupts so far (which already was the case before, but with broken syntax).


device_backends/NumericAddressedBackend/include/NumericAddressedBackend.h
```diff
@@ -174,12 +174,12 @@ namespace ChimeraTK {
      * modified. This map is filled in the constructor. The rest of the code is accessing it through the const
      * _interruptDispatchers reference, which is thread safe.
      */
-    std::map<std::pair<int, int>, boost::shared_ptr<NumericAddressedInterruptDispatcher>> _interruptDispatchersNonConst;
+    std::map<uint32_t, boost::shared_ptr<NumericAddressedInterruptDispatcher>> _primaryInterruptDispatchersNonConst;
```
mhier, line 177: A more elegant (and safer) way would be to store only a const version of the map (not a reference, make the entire map const) and fill it in the constructor initialiser list (via a function call there). This would prevent any modification of the map at a later point.

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
```diff
@@ -118,15 +119,29 @@ namespace ChimeraTK {
...
+      auto getNestedInterruptDispatcher =
+          [](std::vector<uint32_t> interruptID,
+              std::map<uint32_t, boost::shared_ptr<NumericAddressedInterruptDispatcher>> dispatchers,
+              [[maybe_unused]] auto&& _getNestedInterruptDispatcher)
+          -> boost::shared_ptr<NumericAddressedInterruptDispatcher> {
+        auto dispatcher = dispatchers[interruptID.front()];
+        if(interruptID.size() == 1) {
+          return dispatcher;
+        }
+        throw ChimeraTK::logic_error("Nested interrupts are not supported yet!");
```
mhier, line 131: Why do you introduce this lambda if it anyway does not yet recurse?

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
```diff
@@ -216,7 +231,7 @@ namespace ChimeraTK {
       throw ChimeraTK::runtime_error("NumericAddressedBackend is in exception state.");
     }
     catch(...) {
-      for(const auto& it : _interruptDispatchers) {
+      for(const auto& it : _primaryInterruptDispatchers) {
```
mhier, line 234: What about the secondary interrupts? They also need to receive the exception? Are exceptions always propagated to all output interrupts of an interrupt controller?

device_backends/NumericAddressedBackend/src/MapFileParser.cpp
```diff
@@ -109,43 +108,37 @@ namespace ChimeraTK {
...
-    unsigned int interruptCtrlNumber = 0;
-    unsigned int interruptNumber = 0;
     accessTypeStr.erase(pos, strToFind.length());
 
-    auto delimiterPos = accessTypeStr.find(':');
-    if(delimiterPos != std::string::npos) {
-      std::string interruptCtrlNumberStr = accessTypeStr.substr(0, delimiterPos);
-      std::string interrupNumberStr = accessTypeStr.substr(delimiterPos + 1);
+    size_t delimiterPos;
+    do {
```
phako, line 120: I think you can do that with fewer string copying by using from_chars and using the old and new index, but that is probably nitpicking

phako (approved):

device_backends/NumericAddressedBackend/src/MapFileParser.cpp
```diff
@@ -109,43 +108,37 @@ namespace ChimeraTK {
 
   /********************************************************************************************************************/
 
-  std::pair<bool, std::pair<unsigned int, unsigned int>> MapFileParser::getInterruptData(std::string accessTypeStr) {
+  std::vector<uint32_t> MapFileParser::getInterruptData(std::string accessTypeStr) {
```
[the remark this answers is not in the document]
killenb: Ah, that's a very good point. I will refactor that.

device_backends/NumericAddressedBackend/include/NumericAddressedBackend.h
killenb: I don't have this information in the initialiser list. I have to parse and interpret the map file to fill it.
I could make it mutable, but this would not prevent internal functions to modify it, which is the case if they use the const reference.

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
killenb: Yes, you are right. Forget about this part. This has already been refactored in the real implementation.

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
killenb: Yes, they have to, because the all belong to the same device. All the AsyncVariables of all interrupt controllers, no matter how deep they are nested, must see the exception according to spec.

device_backends/NumericAddressedBackend/include/NumericAddressedBackend.h
killenb: Ah, you wrote via function call. I will consider that for the next PR. I has been like that, and this PR is about minimal changes to get the new map file syntax in before the FW teams starts implementing it in RDL.

killenb: I will create a ticket.

device_backends/NumericAddressedBackend/src/MapFileParser.cpp
killenb: I made a comment in the code so it can be improved when this is touched again.

PR #348, opened 2023-05-15
Fix: remove  async write accessor

vaeng:


tests/executables_src/testAsyncNDRegisterAccessor.cpp
killenb, line 1: The code comment already says that most of the functionality is tested with the unified backend test. Now that the write functionality is gone I think there is nothing left. Only the writing had to be tested separately because in the NumericAddressedBackends (i.e. Dummy which is used for testing), the catalogue made the INTERRUPT registers read-only, and this test was working around it.

This test can just be removed.

tests/executables_src/testAsyncNDRegisterAccessor.cpp
```diff
@@ -41,8 +37,8 @@ BOOST_AUTO_TEST_CASE(Test_setWriteAccessor) {
 
   BOOST_CHECK(int(asyncAccessor) != 43);
 
-  asyncAccessor = 43;
-  asyncAccessor.write();
+  syncAccessor = 43;
+  syncAccessor.write();
 
   BOOST_CHECK_EQUAL(d.read<int>("MODULE0/INTERRUPT_TYPE"), 43);
 }
```
killenb, line 44: These lines were the actual test. By changing it to symcAccessor you are only testing that DUMMY_WRITEABLE is working.

killenb (approved):

PR #350, opened 2023-05-16
fix: math plugin and non-redirectedRegister target

mhier:


d-rothe (approved):

d-rothe (commented):
Oh, I'm just seeing testLMapMathPluginPushPars fails now:
error: in "LMapMathPluginTestSuite/testPushPars": check DummyForCleanupCheck::cleanupCalled has failed
This is probably because of the target accessor now being in the math plugin itself. Does it hold the LNM Backend as exception backend? That would mean here's a loop.

mhier:
> Oh, I'm just seeing testLMapMathPluginPushPars fails now: error: in "LMapMathPluginTestSuite/testPushPars": check DummyForCleanupCheck::cleanupCalled has failed This is probably because of the target accessor now being in the math plugin itself. Does it hold the LNM Backend as exception backend? That would mean here's a loop.

Thanks for catching this. I think I have a solution. I can (re-)create the correct accessor in the helper rather than keeping it in the plugin forever with a small modification in the plugin base class.

d-rothe (approved):

PR #354, opened 2023-06-12
Fix 11910 lmap push pars

d-rothe:


device_backends/LogicalNameMapping/include/LNMAccessorPlugin.h
```diff
@@ -71,6 +71,14 @@ namespace ChimeraTK::LNMBackend {
      */
     virtual void openHook(const boost::shared_ptr<LogicalNameMappingBackend>& backend) { std::ignore = backend; }
 
+    /**
+     *  Hook called after the parsing of logical name map. This can be used for setup code which needs complete
+     *  logical name map definitions but must be executed before any register accessor is created.
+     */
+    virtual void postParsingHook(const boost::shared_ptr<const LogicalNameMappingBackend>& backend) {
+      std::ignore = backend;
```
mhier, line 79: I suggest to use `[[maybe_unused]]` instead, which I think can help optimisations. As far as I know, `std::ignore` is intended for use in tuples.

```
virtual void postParsingHook(const boost::shared_ptr<const LogicalNameMappingBackend>& [[maybe_unused]] backend) {
}

mhier (approved):

PR #355, opened 2023-06-13
feat: writeIfDifferent accounts for DataValidity

mhier:


tests/include/WriteCountingBackend.h
```diff
@@ -0,0 +1,35 @@
+// SPDX-FileCopyrightText: Deutsches Elektronen-Synchrotron DESY, MSK, ChimeraTK Project <chimeratk-support@desy.de>
+// SPDX-License-Identifier: LGPL-3.0-or-later
+
```
phako, line 3: I think this fails the code style check without `#pragma once`

phako (approved):

tests/include/WriteCountingBackend.h
mhier: good point, thanks

PR #358, opened 2023-07-25
Unify XDMA and UIO MMIO access

phako:


phako:
I cannot currently test that this didn't break anything on XDMA

patrislav1:
> I cannot currently test that this didn't break anything on XDMA

I should be able to test XDMA with the DAMC-FMC2ZUP example design some time.

patrislav1:
> > I cannot currently test that this didn't break anything on XDMA
> 
> I should be able to test XDMA with the DAMC-FMC2ZUP example design some time.

My [XDMA test program](https://msktechvcs.desy.de/huesmann/xdma-test), which uses BAR 0, BAR 13 (DMA) and interrupts, completed without issues.

device_backends/mmio/include/DeviceFile.h
```diff
@@ -11,18 +11,22 @@ namespace ChimeraTK {
   // Simple RAII wrapper for a device file
   class DeviceFile : private boost::noncopyable {
    private:
-    std::string _strerror(const std::string& msg) const;
-    const std::string _path;
-    int _fd;
+    [[nodiscard]] std::string _strerror(const std::string& msg) const;
```
mhier, line 14: Our coding guide says public comes first and private last...

device_backends/mmio/include/DeviceFile.h
```diff
@@ -11,18 +11,22 @@ namespace ChimeraTK {
...
-    DeviceFile(const std::string& filePath, int flags);
-    DeviceFile(DeviceFile&& d);
+    DeviceFile(std::string filePath, int flags);
+    DeviceFile(int fd, bool takeFdOwnership = true);
+    DeviceFile(DeviceFile&& d) noexcept;
     virtual ~DeviceFile();
 
-    operator int() const;
-    std::string name() const;
+    explicit operator int() const;
```
mhier, line 27: Please add some documentation. What the conversion to int does for a DeviceFile is really not obvious...

device_backends/mmio/include/MmioAccess.h
```diff
@@ -0,0 +1,39 @@
...
+    ~MmioAccess();
+
+    void read(uintptr_t address, int32_t* __restrict__ buf, size_t nBytes);
+    void write(uintptr_t address, const int32_t* data, size_t nBytes);
+
+   private:
+    DeviceFile _file;
+    void* _mem{nullptr};
+
+    // Size of mmap'ed area in PCI BAR
```
mhier, line 26: Is the comment appropriate for UIO as well? It's not PCI, right? Or is this variable only used for xdma?

device_backends/mmio/include/MmioAccess.h
```diff
@@ -0,0 +1,39 @@
...
+    size_t _mmapSize{0};
+
+    // 4 KiB is the minimum size available in Vivado
+    static constexpr size_t _mmapSizeMin = 4UL * 1024UL;
+    static constexpr size_t _mmapSizeMax = 16UL * 1024UL * 1024UL;
```
mhier, line 31: What happens if more than 16MB should be accessed?

device_backends/mmio/include/MmioAccess.h
```diff
@@ -0,0 +1,39 @@
...
+
+    [[nodiscard]] volatile int32_t* _reg_ptr(uintptr_t offs) const;
+    void _check_range(const std::string& access_type, uintptr_t address, size_t nBytes) const;
+    void mapFile(size_t mmapSize);
```
mhier, line 35: Please find a different name for this function. We use the term map file for a file containing a map, here "map" is probably used as a verb. Maybe "mapDeviceFile" would already be more clear.

device_backends/mmio/src/DeviceFile.cc
```diff
@@ -0,0 +1,65 @@
...
+#include <cstring>
+#include <fcntl.h>
+#include <iostream>
+#include <unistd.h>
+#include <utility>
+
+namespace ChimeraTK {
+
+  DeviceFile::DeviceFile(std::string deviceFilePath, int flags) : _path{std::move(deviceFilePath)} {
+#ifdef _DEBUG
```
mhier, line 18: Isn't `_DEBUG` defined for relWithDebInfo builds?

device_backends/mmio/src/DeviceFile.cc
```diff
@@ -0,0 +1,65 @@
...
+    auto path = "/proc/self/fd/" + std::to_string(fd);
+    char buf[PATH_MAX + 1] = {0};
+    auto result = readlink(path.c_str(), buf, sizeof(buf));
+    _savedErrno = errno;
+    if(result < 0) {
+      throw runtime_error(_strerror("Cannot determine path of fd "));
+    }
+
+    _path = std::string{buf};
+    std::cout << "Path for " << _fd << ": " << _path << std::endl;
```
mhier, line 38: Remove debug output...

device_backends/mmio/include/MmioAccess.h
```diff
@@ -0,0 +1,39 @@
...
+    void write(uintptr_t address, const int32_t* data, size_t nBytes);
+
+   private:
+    DeviceFile _file;
+    void* _mem{nullptr};
+
+    // Size of mmap'ed area in PCI BAR
+    size_t _mmapSize{0};
+
+    // 4 KiB is the minimum size available in Vivado
```
mhier, line 29: Isn't the UIO stuff in principle independent of Vivado?

device_backends/uio/src/UioAccess.cc
```diff
@@ -9,14 +9,16 @@
...
 
 namespace ChimeraTK {
 
   UioAccess::UioAccess(const std::string& deviceFilePath) : _deviceFilePath(deviceFilePath.c_str()) {
     std::string fileName = _deviceFilePath.filename().string();
-    _deviceKernelBase = (void*)readUint64HexFromFile("/sys/class/uio/" + fileName + "/maps/map0/addr");
+    //NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+    _deviceKernelBase = reinterpret_cast<void*>(readUint64HexFromFile("/sys/class/uio/" + fileName + "/maps/map0/addr"));
     _deviceMemSize = readUint64HexFromFile("/sys/class/uio/" + fileName + "/maps/map0/size");
     _lastInterruptCount = readUint32FromFile("/sys/class/uio/" + fileName + "/event");
```
mhier, line 24: I know this is not your code, but in line 28 a `runtime_error` is thrown, which should not be done in the constructor of the backend. When does this code run in the end? And why isn't the DeviceFile class used here?

device_backends/uio/src/UioAccess.cc
```diff
@@ -32,14 +34,13 @@ namespace ChimeraTK {
   }
 
   void UioAccess::open() {
-    UioMMap();
+    _mmio = std::make_unique<MmioAccess>(_deviceFileDescriptor, _deviceMemSize, true);
```
mhier, line 37: Why is the MmioAccesso class designed such that you have to use a unique_ptr? Wouldn't it make more sense to use a plain instance and call some init function here?

device_backends/uio/src/UioAccess.cc
```diff
@@ -32,14 +34,13 @@ namespace ChimeraTK {
...
     _opened = true;
   }
 
   void UioAccess::close() {
     if(!_opened) {
-      UioUnmap();
-      ::close(_deviceFileDescriptor);
+      _mmio.reset();
```
mhier, line 43: Here is probably a behaviour change as you removed the `::close(_deviceFileDescriptor)`. Is that intentional?

device_backends/mmio/include/DeviceFile.h
phako: Removed. It was missing the implementation anyway because I decided for the explicit `fd()` call later.

device_backends/mmio/include/DeviceFile.h
phako: it is public first/private last?

device_backends/mmio/include/MmioAccess.h
phako: This comment is wrong. the variable is used in both, though. Fixed the comment

device_backends/mmio/include/MmioAccess.h
phako: This is used for the "autodetection of mmap size" in XDMA. I have modified the name of the variable and the comment

device_backends/mmio/include/MmioAccess.h
phako: I have changed the name of those variables to include "Autodetect" since that is what we support there.

device_backends/mmio/include/MmioAccess.h
phako: renamed to "doMemoryMapping"

device_backends/mmio/src/DeviceFile.cc
phako: Done

device_backends/uio/src/UioAccess.cc
phako: Yeah that make stuff easier I suppose. will change that. That would also move the runtime error into open

device_backends/uio/src/UioAccess.cc
phako: TBH I don't remember

device_backends/uio/src/UioAccess.cc
phako: No, the MmioAccess takes ownership of the file descriptor so resetting the pointer will close the fd and leave it dangling as the old code would.

device_backends/uio/src/UioAccess.cc
mhier: Still, I find it a bit strange to make an object residing in dynamic memory so you can call the constructor and destructor at defined points in time, and then pack the dynamic object into some container to make sure it does not leak.

If you had separate member functions to control the required actions at open and close and just call close in the destructor you could get away without dynamic memory allocation and containers. RAII is a nice concept in other context, but here you merely want to split off certain code into its own class. I don't see the benefit of having a different lifetime of the new object than its parent.

device_backends/uio/src/UioAccess.cc
mhier: Uh, that's not clear from the code. Can the MmioAccess object even create the _deviceFileDescriptor in the first place so the parent class does not deal with it at all? I guess that would be inline with some of the previous comments...

phako: This will need deeper rework since it affects the interrupt handler, I am already looking into it. I think it will be much clearer in the end

patrislav1: I would prefer using [DeviceFile](https://github.com/ChimeraTK/DeviceAccess/blob/master/device_backends/xdma/include/DeviceFile.h) for that, instead of dealing with explicit fd's and explicit closing, but looking at https://github.com/ChimeraTK/DeviceAccess/pull/358#discussion_r1312699597 maybe that doesn't fit into the style of the rest of the framework.

mhier: That comment was not against the DeviceFile object in total, merely a small question mark about a design aspect of DeviceFile...

phako:
We will do something different here

PR #371, opened 2023-11-22
fix: Memcopy for mmio is bad, do manual copy

phako:
e.g. on ARM64, memcpy will result in more than 1 write to an address, causing multiple triggers to the FPGA for this data to happen

### DIFF
--- a/device_backends/uio/src/UioAccess.cc
+++ b/device_backends/uio/src/UioAccess.cc
@@ -44,7 +44,7 @@ namespace ChimeraTK {
     }
   }
 
-  void UioAccess::read(uint64_t map, uint64_t address, int32_t* data, size_t sizeInBytes) {
+  void UioAccess::read(uint64_t map, uint64_t address, int32_t* __restrict__ data, size_t sizeInBytes) {
     if(map > 0) {
       throw ChimeraTK::logic_error("UIO: Multiple memory regions are not supported");
     }
@@ -56,10 +56,11 @@ namespace ChimeraTK {
       throw ChimeraTK::logic_error("UIO: Read request exceeds device memory region");
     }
 
-    void* targetAddress = static_cast<uint8_t*>(_deviceUserBase) + address;
-
-    // Is inherently thread-safe
-    std::memcpy(data, targetAddress, sizeInBytes);
+    volatile int32_t* rptr = static_cast<volatile int32_t*>(_deviceUserBase) + address / sizeof(int32_t);
+    while(sizeInBytes >= sizeof(int32_t)) {
+      *data++ = *rptr++;
+      sizeInBytes -= sizeof(int32_t);
+    }
   }
 
   void UioAccess::write(uint64_t map, uint64_t address, int32_t const* data, size_t sizeInBytes) {
@@ -73,10 +74,12 @@ namespace ChimeraTK {
     if(address + sizeInBytes > _deviceMemSize) {
       throw ChimeraTK::logic_error("UIO: Write request exceeds device memory region");
     }
-    void* targetAddress = static_cast<uint8_t*>(_deviceUserBase) + address;
 
-    // Is inherently thread-safe
-    std::memcpy(targetAddress, data, sizeInBytes);
+    volatile int32_t* __restrict__ wptr = static_cast<volatile int32_t*>(_deviceUserBase) + address / sizeof(int32_t);
+    while(sizeInBytes >= sizeof(int32_t)) {
+      *wptr++ = *data++;
+      sizeInBytes -= sizeof(int32_t);
+    }
   }
 
   uint32_t UioAccess::waitForInterrupt(int timeoutMs) {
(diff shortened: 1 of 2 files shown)


phako:
This is the "replacement" for #358 which just fixes the memcpy issue

device_backends/uio/src/UioAccess.cc
```diff
@@ -56,10 +56,11 @@ namespace ChimeraTK {
       throw ChimeraTK::logic_error("UIO: Read request exceeds device memory region");
     }
 
-    void* targetAddress = static_cast<uint8_t*>(_deviceUserBase) + address;
-
-    // Is inherently thread-safe
-    std::memcpy(data, targetAddress, sizeInBytes);
+    volatile int32_t* rptr = static_cast<volatile int32_t*>(_deviceUserBase) + address / sizeof(int32_t);
+    while(sizeInBytes >= sizeof(int32_t)) {
+      *data++ = *rptr++;
```
killenb, line 61: I never know which operator has precedence. I guess this is ```*(data++) = *(rptr++)```?

killenb: Please make it explicit.

device_backends/uio/src/UioAccess.cc
```diff
@@ -56,10 +56,11 @@ namespace ChimeraTK {
       throw ChimeraTK::logic_error("UIO: Read request exceeds device memory region");
     }
 
-    void* targetAddress = static_cast<uint8_t*>(_deviceUserBase) + address;
-
-    // Is inherently thread-safe
-    std::memcpy(data, targetAddress, sizeInBytes);
+    volatile int32_t* rptr = static_cast<volatile int32_t*>(_deviceUserBase) + address / sizeof(int32_t);
+    while(sizeInBytes >= sizeof(int32_t)) {
```
killenb, line 60: There are up to three bytes uncopied. I guess this is not a valid use case with UIO? Do we have a check for it?

device_backends/uio/src/UioAccess.cc
```diff
@@ -73,10 +74,12 @@ namespace ChimeraTK {
     if(address + sizeInBytes > _deviceMemSize) {
       throw ChimeraTK::logic_error("UIO: Write request exceeds device memory region");
     }
-    void* targetAddress = static_cast<uint8_t*>(_deviceUserBase) + address;
 
-    // Is inherently thread-safe
-    std::memcpy(targetAddress, data, sizeInBytes);
+    volatile int32_t* __restrict__ wptr = static_cast<volatile int32_t*>(_deviceUserBase) + address / sizeof(int32_t);
+    while(sizeInBytes >= sizeof(int32_t)) {
```
killenb, line 79: see above

device_backends/uio/src/UioAccess.cc
```diff
@@ -73,10 +74,12 @@ namespace ChimeraTK {
...
+      *wptr++ = *data++;
```
killenb, line 80: see above

device_backends/uio/src/UioAccess.cc
phako: If you do that, it first increments, then assigns, skipping the first entry.

will do
```
*data = *rptr
data++; rptr++
```

device_backends/uio/src/UioAccess.cc
phako: Good question, for our UIO this is correct, probably not for the generic case?

device_backends/uio/src/UioAccess.cc
phako: See above

killenb:
> I just thought that this will also be problematic if we do not start on a multiple of 4?

That's a good point. And wasn't this writing in chunks of 32 to prevent memcopy from writing in 64 bit blocks? What happens on a ARM 32 if you write an uneven number of shorts? Ah no, we write the remainder byte by byte.

phako:
We cannot do copy byte-by-byte, that would be 4 writes per int32 register and we would be back to square 1

killenb:
But what do we do with the remaining not multiple of 4 bytes? We have to write them bye by byte, so it is up to 3 writes.

This is special because it assumes we have an underying  32 bit layout, and having those "trigger registers" aligned with it. We somehow need a way to configure that. Or don't we? Somehow the requirements are not clear.

We could also just check that the size is a multiple of 4 bytes and throw if it is not. But this makes the backed less universal.

mhier:
> We could also just check that the size is a multiple of 4 bytes and throw if it is not. But this makes the backed less universal.

You can simply overload `NumericAddressedBackend::minimumTransferAlignment()` to return 4, in which case there will never be done any transfer which is not aligned at integer multiple of 4 bytes. Smaller registers will be accessed by issuing a wider transfer and throwing away bits (resp. read-modify-write for the write direction). This is how we have it for the pcie backend etc.

PR #376, opened 2023-11-30
fix(xdma): event files opened in the wrong place

killenb:
- push type accessors can be requested before opening the device
- no event file is opened if no push type accessors exist

### DIFF
--- a/device_backends/xdma/include/XdmaBackend.h
+++ b/device_backends/xdma/include/XdmaBackend.h
@@ -22,7 +22,8 @@ namespace ChimeraTK {
 
     std::optional<CtrlIntf> _ctrlIntf;
     std::vector<DmaIntf> _dmaChannels;
-    std::vector<EventFile> _eventFiles;
+    std::array<std::unique_ptr<EventFile>, _maxInterrupts> _eventFiles;
+    std::array<bool, _maxInterrupts> _startInterruptHandlingCalled{};
 
     const std::string _devicePath;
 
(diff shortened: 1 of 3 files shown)


d-rothe (approved):
Test confirms it now runs with ApplicationCore (which first obtains RegisterAccessors, then opens device)

device_backends/xdma/include/XdmaBackend.h
```diff
@@ -22,7 +22,8 @@ namespace ChimeraTK {
 
     std::optional<CtrlIntf> _ctrlIntf;
     std::vector<DmaIntf> _dmaChannels;
-    std::vector<EventFile> _eventFiles;
+    std::array<std::unique_ptr<EventFile>, _maxInterrupts> _eventFiles;
+    std::array<bool, _maxInterrupts> _interruptThreadStarted{};
```
phako, line 27: Whenever I see two arrays with the same length with data handled together - Shouldn't that be one array with a struct (or pair in this case)?

killenb: In principle I agree. However, you either have to define the struct (more lines of code), or in case of a pair the usage becomes (I don't even have a name for the combined thing) _interruptSometings[i].first and _interruptSometings[i].second, which is ugly to read. I usually put both into the name to know what first and second is:  _eventFilesAndIsActiveFlags[i].second is the flag. I don't know if that's better style???

PR #377, opened 2023-12-22
hotfix: rethrow old exceptions by math plugin after recovery

mhier:
Note: This change is likely to be replaced with a better implementation (with potential side-effects, so not to be released as a patch).

The problem solved with this commit is that the math plugin with push parameters prevented the recovery from a runtime_error, because the old runtime_error was still present in the parameter queues after re-opening the backend and was hence re-thrown in the next write operation, placing the backend again into the exception state. This is now solved by clearing the queues in the open_hook.

A better implementation might be to not use wait_for_new_data for the parameters (even if push parameters are enabled), so there are no queues at all which could store the exception. wait_for_new_data is in principle not necessary, since the math plugin never executes blocking reads on the paramters. Instead the push parameter variables will trigger the update of the math plugin in its postWrite directly. As a side effect, the math plugin push parameters would work independently from activateAsyncRead(), which probably can be considered cleaner and more intuitive.


device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -142,6 +147,21 @@ namespace ChimeraTK::LNMBackend {
 
   /********************************************************************************************************************/
 
+  void MathPluginFormulaHelper::clearParameterQueues() {
+    for(auto& p : params) {
+      bool rv = true;
```
killenb, line 152: What is rv? Although this is a very local variable I cannot guess the name. Is it "there has been data on the queue"?

device_backends/LogicalNameMapping/src/LNMMathPlugin.cc
```diff
@@ -77,6 +77,11 @@ namespace ChimeraTK::LNMBackend {
       _mainValueWrittenAfterOpen = false;
       _allParametersWrittenAfterOpen = false;
     }
+
+    // consume all entries from the push-type parameter queues, to prevent old exceptions from being thrown again
+    if(_hasPushParameter) {
+      getFormulaHelper(backend)->clearParameterQueues();
```
killenb, line 83: When you first showed me this you mentioned that this might unnecessarily instantiate a formula helper. Do you think it's worth putting this information into a comment? With the improved solution without queues this section will go out anyway.

killenb (approved):
I made two comments, which both are minor. You can either implement them or just close the conversation and merge it like it is.

PR #378, opened 2023-12-22
feat!(exceptions): keep and report original message

mhier:


device_backends/src/DummyInterruptTriggerAccessor.cc
```diff
@@ -3,6 +3,86 @@
...
+  bool DummyInterruptTriggerAccessor<UserType>::doWriteTransfer(ChimeraTK::VersionNumber) {
+    _interruptTrigger();
+
+    return false;
+  }
+
+  /********************************************************************************************************************/
+
+  template<typename UserType>
+  void DummyInterruptTriggerAccessor<UserType>::doPreRead(TransferType) {
```
killenb, line 47: The implementation was like that, but conceptually this accessor is void and write-only (if I am not mistaken). It does not make sense to implement the read-functions at all.

device_backends/uio/src/UioBackend.cc
```diff
@@ -108,14 +96,14 @@ namespace ChimeraTK {
...
-        setException();
+        setException(ex.what());
         continue;
       }
 
       if(numberOfInterrupts > 0) {
         _uioAccess->clearInterrupts();
 
-        if(_hasActiveException) {
+        if(isFunctional()) {
```
killenb, line 106: There is a NOT missing:
```suggestion
        if(!isFunctional()) {
```

tests/executables_src/testDevice.cpp
```diff
@@ -176,13 +167,6 @@ BOOST_AUTO_TEST_CASE(testIsFunctional) {
 
   d.close();
   BOOST_CHECK(d.isFunctional() == false);
-
-  ChimeraTK::BackendFactory::getInstance().registerBackendType("DysfunctDummy", &DysfunctDummy::createInstance);
```
killenb, line 180: Removing this tests is good because I think open() on the dysfunct dummy should have thrown. I guess that the device is still open after throwing an exception is part of the unified backend test. To test it here make no sense because it is implementation specific.

tests/executables_src/testSubdeviceBackend.cpp
```diff
@@ -692,10 +692,10 @@ BOOST_AUTO_TEST_CASE(testIsFunctional) {
   target.open("TARGET1");
   BOOST_CHECK(dev.isFunctional());
   // Close target device.
```
killenb, line 694: The comment does not match, and provides no added value anyway.
```suggestion
```

killenb (approved):
setOpenedAndClearException() in DeviceBackendImpl should also go to the .cc file. And there is a bug (missing NOT) in one of the backend implementations. If those are addressed the change can be merged.

device_backends/DummyBackend/src/ExceptionDummyBackend.cc
```diff
@@ -103,26 +103,26 @@ namespace ChimeraTK {
...
     }
     ChimeraTK::DummyBackend::open();
-    thereHaveBeenExceptions = false;
   }
 
   /********************************************************************************************************************/
 
   void ExceptionDummy::closeImpl() {
-    setException();
+    setException("Close ExceptionDummy");
```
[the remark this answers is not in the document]
mhier: setException does not throw. I think this line is completely redundant, since the exception state cannot be seen from the outside. The backend is closed, hence any operation would lead to a logic_error. (Otherwise the spec is violated)

device_backends/include/DeviceBackendImpl.h
```diff
@@ -28,11 +29,97 @@ namespace ChimeraTK {
...
+     */
+    std::string _activeExceptionMessage;
+
+    /** mutex to protect access to _activeExceptionMessage */
+    std::mutex _mx_activeExceptionMessage;
   };
 
   /********************************************************************************************************************/
 
+  /// TODO move to .cc file
```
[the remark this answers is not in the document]
mhier: Sorry, I noticed yesterday that I forgot to do that and already pushed a fix fir this.

device_backends/DummyBackend/include/DummyBackendBase.h
```diff
@@ -30,7 +30,7 @@ namespace ChimeraTK {
    * Base class for DummyBackends, provides common functionality
    */
   class DummyBackendBase : public NumericAddressedBackend {
-   protected:
+   public:
     // ctor & dtor private with derived type as friend to enforce correct specialization
```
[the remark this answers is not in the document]
mhier:
```suggestion
```

device_backends/DummyBackend/src/ExceptionDummyBackend.cc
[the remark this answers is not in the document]
mhier:
```suggestion
```

device_backends/src/DummyInterruptTriggerAccessor.cc
mhier: I think we should do that separately. (Currently, I don't have the overview what this exactly does.)

mhier: See redmine issue #12628

PR #379, opened 2023-12-22
fix: rethrow old exceptions by math plugin after recovery

mhier:
This replaces the earlier hotfix with a better implementation.

As a (positive) side effect of the new fix, push-type parameters of math plugins now work independently of the async read activation state.

### DIFF
--- a/device_backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
+++ b/device_backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
@@ -232,8 +232,12 @@ namespace ChimeraTK {
         auto& vtEntry = boost::fusion::at_key<decltype(arg)>(lnmVariable.valueTable.table);
         for(size_t i = 0; i < this->buffer_2D[0].size(); ++i) {
           this->buffer_2D[0][i] = userTypeToUserType<UserType>(vtEntry.latestValue[i + _wordOffsetInRegister]);
-          this->_dataValidity = vtEntry.latestValidity;
         }
+        this->_dataValidity = vtEntry.latestValidity;
+        // Note: passing through the version number also for push-type variables is essential for the MathPlugin (cf.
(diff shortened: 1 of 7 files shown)


d-rothe (approved):

device_backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
```diff
@@ -234,8 +234,8 @@ namespace ChimeraTK {
           this->buffer_2D[0][i] = userTypeToUserType<UserType>(vtEntry.latestValue[i + _wordOffsetInRegister]);
           this->_dataValidity = vtEntry.latestValidity;
```
d-rothe, line 235: this could be move out of the loop (one line down)

device_backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
```diff
@@ -234,8 +234,8 @@ namespace ChimeraTK {
...
         }
+        this->_versionNumber = vtEntry.latestVersion;
```
d-rothe, line 237: question for my understanding: is it actually necessary for the new math-plugin behavior, to keep the latest version number even for poll-type acessors (on the parameters), or is this an independent improvement you just added here?

mhier: It is essential, since otherwise the check in `MathPluginFormulaHelper::checkAllParametersWritten()` no longer works, because we now obtain even push-type parameters without wait_for_new_data. The check would always evaluate true and garbage values would be written to the device in the startup phase.

d-rothe: I see. Can you please update the comment in MathPluginFormulaHelper::checkAllParametersWritten since it's no longer correct.

mhier: done. I also added a comment here explaining that it is important for the math plugin and that it does not violate the spec.

PR #391, opened 2024-02-19
Feat/new version after open test

mhier:


killenb (approved):
Will there be a unit test for the math plugin with constants and variables? The current test is not enough for the math plugin because the "version on open" would be sufficient.

mhier:
> Will there be a unit test for the math plugin with constants and variables? The current test is not enough for the math plugin because the "version on open" would be sufficient.

I created a follow up ticket for this now: https://redmine.msktools.desy.de/issues/12833

PR #393, opened 2024-02-20
fix: structured bind in lambdas

killenb:
Clang does not support the use of variables from a structured bind
in lambdas yet.


mhier (approved):
upps, thanks!

PR #402, opened 2024-03-11
fix(xdma): open event file only when you acquire an accessor

nshehz:


device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -39,15 +39,8 @@ namespace ChimeraTK {
       }
     }
 
-    // (Re-)Open the event files which are needed
+    // (Re-)Open the event files only when you acquire an accessaor.
     std::for_each(_eventFiles.begin(), _eventFiles.end(), [](auto& eventFile) { eventFile = nullptr; });
```
killenb, line 43: These the comment is not quite correct. You open the event file when startInterruptHandlingThread is called (the Thread lives inside the EventFile class).
You can remove both the comment and the code line. Cleaning in open() is not useful. It is happening close() and probably should happen in setException (to be discussed), but open is definitely the wrong place.
```suggestion
```

device_backends/xdma/src/XdmaBackend.cc
```diff
@@ -39,15 +39,8 @@ namespace ChimeraTK {
...
-    for(size_t i = 0; i < _maxInterrupts; i++) {
-      if(_startInterruptHandlingCalled[i]) {
```
killenb, line 45: With this code gone, the variable _startInterruptHandlingCalled is obsolete. Please remove it.

killenb (approved):
You proceed with the merge once the two conversations are addressed.

device_backends/xdma/src/XdmaBackend.cc
nshehz: done

PR #410, opened 2024-03-28
fix(NumericAddressedBackend): AsyncDomain weak pointer locking

killenb:
The same instance of a weak pointer cannot be used/locked by multiple threads.
- Use AsyncDomainsContainer with its lock to lock weak pointers
- Use separate pointer instances to avoid the lock for trigger distribution


device_backends/include/AsyncDomainImpl.h
```diff
@@ -47,7 +47,8 @@ namespace ChimeraTK {
      *  @ return The version number that has been used for distribution.
      */
     VersionNumber activate(BackendDataType data, VersionNumber version = VersionNumber{nullptr});
-    void deactivate();
+
+    void deactivate() override;
     void sendException(const std::exception_ptr& e) noexcept override;
 
     template<typename UserDataType>
```
mhier, line 54: `subscribe()` could become a virtual function as well (with the ususal trick for UserType-templated virtual functions), which might be helpful for backends that have varying `BackendDataType`. For those backends (maybe even generally for all backends) I would implement some kind of "createOrGetAsyncDomain" function, in which case the calling code (e.g. `getRegisterAccessor()`) does not need to know the backend type.

device_backends/include/AsyncDomainsContainer.h
```diff
@@ -18,126 +16,70 @@ namespace ChimeraTK {
...
+    /** Check whether an exception distribution is started and not completed yet. */
+    bool isSendingExceptions() { return _isSendingExceptions; }
 
     /**
      * Adds an AyncDomain to the container. If there is already an entry for that key, and that weak pointer can be
      * locked, it throws a ChimeraTK::logic_error. This function must only be called if the either an entry for that key
      * does not exist, or the according share pointer is gone.
      */
-    void addAsyncDomain(KeyType key, const boost::shared_ptr<AsyncDomain>& asyncDomain);
+    void addAsyncDomain(size_t key, const boost::shared_ptr<AsyncDomain>& asyncDomain);
```
mhier, line 52: Maybe to clarify in the description: The `asyncDomain` must be empty when adding, so there must not yet be any subscribers attached to it (otherwise there might be race conditions with exceptions).

I am wondering whether it is possible to write some kind of "getOrCreateAsyncDomain" so we don't need handle these details correctly in the calling code...

device_backends/uio/src/UioBackend.cc
```diff
@@ -6,7 +6,7 @@
 namespace ChimeraTK {
 
   UioBackend::UioBackend(std::string deviceName, std::string mapFileName) : NumericAddressedBackend(mapFileName) {
-    _uioAccess = std::shared_ptr<UioAccess>(new UioAccess("/dev/" + deviceName));
+    _uioAccess = std::make_shared<UioAccess>("/dev/" + deviceName);
```
mhier, line 9: do these changes really belong into the same commit with the other changes?

mhier (approved):

device_backends/SharedDummy/src/SharedDummyBackend.cc
```diff
@@ -309,13 +309,12 @@ namespace ChimeraTK {
...
-    catch(std::out_of_range&) {
-      throw ChimeraTK::logic_error(
-          "InterruptDispatcherThread::triggerInterrupt(): Error: Unknown interrupt " + std::to_string(interruptNumber));
+    SharedDummyBackend& backend = _dispatcherInterf->_backend;
+    auto asyncDomain =
+        boost::dynamic_pointer_cast<VoidAsyncDomain>(backend._asyncDomainsContainer.getAsyncDomain(interruptNumber));
+
+    if(asyncDomain) {
+      asyncDomain->distribute(nullptr);
     }
```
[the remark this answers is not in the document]
killenb: done

device_backends/DummyBackend/src/DummyBackend.cc
```diff
@@ -185,13 +184,13 @@ namespace ChimeraTK {
...
-    catch(std::out_of_range&) {
-      throw ChimeraTK::logic_error(
-          "DummyBackend::triggerInterrupt(): Error: Unknown interrupt " + std::to_string(interruptNumber));
+    auto asyncDomain =
+        boost::dynamic_pointer_cast<VoidAsyncDomain>(_asyncDomainsContainer.getAsyncDomain(interruptNumber));
+
+    if(asyncDomain) {
+      return asyncDomain->distribute(nullptr);
     }
+    return VersionNumber{nullptr};
```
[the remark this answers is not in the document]
killenb: done

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
```diff
@@ -135,19 +126,18 @@ namespace ChimeraTK {
             "Register " + registerPathName + " does not support AccessMode::wait_for_new_data.");
       }
 
-      auto asyncDomain = _asyncDomainImpls.at(registerInfo.interruptId.front())->lock();
+      auto asyncDomain = boost::dynamic_pointer_cast<VoidAsyncDomain>(
+          _asyncDomainsContainer.getAsyncDomain(registerInfo.interruptId.front()));
 
       if(!asyncDomain) {
```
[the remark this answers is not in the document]
killenb: done

device_backends/uio/src/UioBackend.cc
killenb: Well, they are minor improvements in the noise of the changes which are necessary in that file anyway. I did not consider them worth a separate commit.

device_backends/include/AsyncDomainsContainer.h
killenb: The AsyncDomain does not have to be empty because it has a lock itself. If an accessor is already in the AsyncDomain, it will see the exception if the domain was active. If the exception comes first, the domain is inactive and the new accessor does not see an exception. Distributing the exception, creating the accessor and turning on/add the active flag is all happening under the lock inside the AsyncDomain.

However, there is a race condition here: after getAsyncDomain() returns nullptr, a different thread might have added the domain already, so addAsyncDomain() will throw. This violates the specification because it says that you can always check if logic_error will be thrown. However, this is also true for throwing a logic error when a device is closed and you try to read an accessor. Another thread might have closed it after you have just checked.

mhier: > However, this is also true for throwing a logic error when a device is closed and you try to read an accessor. Another thread might have closed it after you have just checked.

You keep repeating this argument for making anything else also not thread-safe. We agreed on `close()` not being thread safe, but everything else must be thread safe.

Anyway: It is a runtime_error, not a logic_error, probably neither is correct. A runtime_error means, the remote device (or the connection to it) does not behave as expected, which is clearly not the case. A logic_error means, the application programmer has done something wrong (which implies that he could have known this beforehand) - in that case at least the message would not be helpful at all, since the application programmer does not even know that there is an AsyncDomain involved.

This needs to be solved differently.

mhier (changes requested):
`runtime_error` if `AsyncDomain` already exists is wrong, also possible race condition when getting register accessors concurrently in different threads needs to be prevented.

device_backends/include/AsyncDomainImpl.h
killenb: I agree that ```subscribe()``` could become a UserType-templated virtual functions. However, debugging those is a pain in the a.... The compiler error messages turn in circles and rarely point to the line that has the actual error, and stepping through with the debugger is very tedious. I would only use them where it is absolutely necessary.

I introduced ```getOrCreateAsyncDomain()``` in the ```AsyncDomainsContainer```, which is returning a shared pointer to the implementation, not to the base class. The function needs to be templated to the backend because it internally has to get the initial value from the backend, which is of the backend specific user type and might vary from domain to domain. The calling code has to provide this information to ``getOrCreateAsyncDomain()``` as template parameter, which is always called before the ```subscribe()``` call, so there is no benefit from adding  UserType-templated virtual functions for ```subscribe()```.

device_backends/include/AsyncDomainsContainer.h
killenb: The function has been replaced with  ```getOrCreateAsyncDomain()```, which creates the async domain under the container lock and removes the race condition.

device_backends/include/DeviceBackend.h
```diff
@@ -96,6 +99,12 @@ namespace ChimeraTK {
      * the appropriate ChimeraTK::runtime_error is thrown by this function.
      */
     virtual void checkActiveException() = 0;
+
+    /**
+     * Create an InterruptControllerHandler for a specific controllerID and a TriggerDistributor as parent.
+     */
+    virtual boost::shared_ptr<InterruptControllerHandler> createInterruptControllerHandler(
```
killenb, line 106: Will be renamed later when InterruptControllerHandler is renamed to MuxedInterruptDistributor.

device_backends/src/TriggerDistributor.cc
```diff
@@ -126,32 +132,38 @@ namespace ChimeraTK {
...
+    template<typename UserType>
+    boost::shared_ptr<AsyncNDRegisterAccessor<UserType>> AsyncDataAdapterSubscriptionImplementor<UserType,
+        std::nullptr_t>::subscribeTo(TriggerDistributor<std::nullptr_t>& triggerDistributor, RegisterPath name,
+        size_t numberOfWords, size_t wordOffsetInRegister, AccessModeFlags flags) {
+      auto catalogue = triggerDistributor._backend->getRegisterCatalogue(); // need to store the clone you get
+      const auto& backendCatalogue = catalogue.getImpl();
+      // This code only works for backends which use the NumericAddressedRegisterCatalogue because we need the
+      // interrupt description which is specific for those backends and not in the general catalogue.
+      // If the cast fails, it will throw an exception.
+      auto removeMe = backendCatalogue.getRegister(name);
```
mhier: What does this line do? `removeMe` isn't referenced anywhere... If it's aiming for a side-effect, maybe add a comment and add `[[maybe_unused]]` to avoid a warning.

device_backends/include/TriggerDistributor.h
```diff
@@ -104,4 +109,138 @@ namespace ChimeraTK {
         *this, name, numberOfWords, wordOffsetInRegister, flags);
   }
 
+  /********************************************************************************************************************/
+
+  template<typename BackendSpecificDataType>
+  template<typename DistributorType>
+  boost::shared_ptr<DistributorType> TriggerDistributor<BackendSpecificDataType>::getDistributorRecursive(
```
mhier, line 108: Maybe rename into `findDistributor()`?

mhier (approved):

device_backends/include/TriggerDistributor.h
killenb: No, it's not internal. The IDE does not find all use cases. I will think of improving the name.

killenb: I will rename this later, together with the other API changes.
Once the QualifiedAsyncId has been introduces and the TriggerDistributor has been re-named, everything will make more sense.
I don't like "find". I don't have to search for it, I know exactly where it is. And "find" often means that it might not be found, which is not the case because it is created if it does not exist.
Probably "getDistributor()" will be a good name once the code reads
```subDomain->getDistributor(qualifiedAsyncId);```
At the moment it would be
```_distributor->getDistributor(interruptId);```, which is confusing (and it is
```_distributor->getDistributorRecursive(interruptId);``` which is slightly less confusing because the recursive gives a hint why you ask something called distributor for a distributor. It is just a mess at the moment.

device_backends/src/TriggerDistributor.cc
killenb: removed

Issue #193, opened 2020-11-13
Move to libxml++3 api (or use another XML library)


phako:

mhier:
Is that available on Ubuntu 20.04? I see only a libxml++2.6-dev package...

phako:
Apparently not. Which is odd, I think it is older than the Ubuntu LTS freeze. but ok.

mhier:
I think it is way older... I cannot find the source code repo for libxml++, the link on the webpage is dead. If it stays like this, we seriously should look for alternatives... Aren't there anyway more C++ish parsers available nowadays?

phako:
Latest release is from January: https://download.gnome.org/sources/libxml++/4.9/ There was some discussion moving the non-GNOME related cpp parts like sigc++ and xml++ somewhere else but I don't know where they went to. there's also 4.0 and 5.0 API. 

https://github.com/libxmlplusplus/libxmlplusplus is the correct one, https://libxmlplusplus.github.io/libxmlplusplus/ the correct page. I suspect the sf page still "belongs" to Murray

phako:
Still trying to talk to people, having libxml++5 api would be nice, since it removes a bit of dependency

PR #411, opened 2024-04-04
Looking for libxml++-5.0 before falling back to libxml++-2.6.

gian21391:
As pointed out in #193, the usage of libxml++-5.0 gives no problems.
I have successfully tested the usage of a statically linked version libxml++-5.0, so why not use it if it is available in the system?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -37,8 +37,13 @@ FIND_PACKAGE(exprtk 01.04 REQUIRED)
 FIND_PACKAGE(nlohmann_json 03.07 REQUIRED)
 
 FIND_PACKAGE(PkgConfig REQUIRED)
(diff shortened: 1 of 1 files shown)


phako:
I am doing that for Yocto currently - which distribution in the wild did you find that actually used libxml++5?

phako:
The slight problem we have is that we would need to support it in all other depending projects as well and that isn't plain sailing for the rest, e.g. https://github.com/ChimeraTK/meta-chimeratk/blob/master/recipes-applicationcore/applicationcore/applicationcore/0001-Bump-libxml-to-5.0.patch and https://github.com/ChimeraTK/meta-chimeratk/blob/master/recipes-applicationcore/controlsystemadapter/controlsystemadapter/0001-Bump-libxml-version-to-5.0.patch

PR #412, opened 2024-04-04
Avoid implicit copy in lambda when only 'this' is needed

gian21391:
The implicit capture of *this, when the capture default is =, is deprecated (since C++20).


phako:
Does this work with 17?

gian21391:
Yes, this works for any version of C++ that has lambda support: https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture

phako:
The slight problem we have is that we would support it in all other depending projects as well and that isn't plain sailing for the rest.

PR #416, opened 2024-04-16
fix failing test

killenb:
- fix: dummy bugs
testDummyBackendUnified was failing since the DummyInterruptTriggerAccessor was correctly reporting that it is not readable.
The failure was due to a bug in the ExceptionDummyBackend. The decorator of the ExceptionDummyBackend was also masking a wrong value
in the user buffer of the DummyInterruptTriggerAccessor.

﻿- fix(UnifiedBackendTest): catalogue test disabled for dummy accessors
The DummyInterruptTriggerAccessor reporting the wrong readability was not detected because the Unified backend test was not sensitive.
The workaround has been removed since the dummy accessors are in the catalogue since some time.

- fix: compiler warning

### DIFF
--- a/device_backends/DummyBackend/include/DummyInterruptTriggerAccessor.h
+++ b/device_backends/DummyBackend/include/DummyInterruptTriggerAccessor.h
@@ -3,7 +3,6 @@
 #pragma once
 
 #include "NDRegisterAccessor.h"
-#include "NumericAddressedBackend.h"
 
 #include <utility>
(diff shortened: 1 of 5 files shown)


toseko:
I recompiled and run tests both for debug and release compilation types. All works. 
I do not have comments to the code. 
All is fine for me.

toseko (approved):
All looks good for me.

PR #417, opened 2024-04-19
chore: improve async API names

killenb:
The aync scheme has evolved and class names have to be improved to match the overall concept and make it understandable.
The namespace async has been added.
These changes only affect the backend-internal async API, which is not used outside of the DeviceAccess core library yet.


device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
```diff
@@ -3,15 +3,15 @@
...
+#include "async/DomainsContainer.h"
+#include "async/SubDomain.h"
 #include "Exception.h"
 #include "MapFileParser.h"
 #include "NumericAddress.h"
 #include "NumericAddressedBackendASCIIAccessor.h"
 #include "NumericAddressedBackendMuxedRegisterAccessor.h"
 #include "NumericAddressedBackendRegisterAccessor.h"
-#include "TriggerDistributor.h"
 #include <nlohmann/json.hpp>
```
mhier, line 15: The json include is no longer required here, I guess?

device_backends/src/async/MuxedInterruptDistributor.cc
```diff
@@ -0,0 +1,126 @@
...
+      if(key[0] == '!') {
+        auto jkey = nlohmann::json::parse(std::string({++key.begin(), key.end()})); // key without the !
+        auto interruptId = jkey.get<std::vector<size_t>>();
+
+        auto jdescriptor = nlohmann::json::parse(metaDataEntry.second);
+        auto controllerType = jdescriptor.begin().key();
+        auto controllerDescription = jdescriptor.front().dump();
+        if(interruptId == subdomainID) {
+          return {controllerType, controllerDescription};
+        }
```
mhier, line 46:
```suggestion
        if(interruptId == subdomainID) {
          auto jdescriptor = nlohmann::json::parse(metaDataEntry.second);
          auto controllerType = jdescriptor.begin().key();
          auto controllerDescription = jdescriptor.front().dump();
          return {controllerType, controllerDescription};
        }
```

But also: You are first parsing the description, then you dump it and later in the create function you parse it again, if I understand the code correctly. Shouldn't we just pass on the parsed object?

mhier (approved):

device_backends/src/async/MuxedInterruptDistributor.cc
killenb: Yes, it has to be in two steps. Here, we just get the type of controller and keep the controller-type specific description unparsed. The factory needs to know the type to know which creator function to call, but it cannot parse the description yet. This is happening in the specific creator function.

device_backends/NumericAddressedBackend/src/NumericAddressedBackend.cc
killenb: removed

device_backends/src/async/MuxedInterruptDistributor.cc
mhier: I was just referring to the JSON parsing, of course the factory can parse the entire JSON string, it already does that. It just cannot interpret the content, but it could pass it on instead of the string.

killenb: Ah, I could have passed on an nlohmann::json object instead of a string, right. This interface is still from the time where the string was coming directly from the catalogue, and there was no json parsing in the first step (if I remember correctly).

mhier: On the other hand: We might want to avoid exposing the nlohmann::json object in the public API, unless we already do so, since it would add another dependency for all applications using DeviceAccess.

killenb: That is a good point. the nlohmann is in none of the headers. I will leave it like this.

PR #418, opened 2024-04-22
fix: include path of headers in async subdir

killenb:


phako:
I think this was too much:

```
/home/jgeorg/Source/workflow-scripts/install-debug/include/ChimeraTK/async/AsyncNDRegisterAccessor.h:5:10: fatal error: DeviceBackend.h: No such file or directory
    5 | #include "DeviceBackend.h"
      |          ^~~~~~~~~~~~~~~~~
compilation terminated.
[3/17] Linking CXX shared l

mhier (approved):

killenb:
I removed the install directory and tried with ControlSystemAdapter and ApplicationCore.
Before the fix the ControlSystemAdapter failed as described by @phako.

phako (commented):
Seems to work after the latest changes.

PR #420, opened 2024-04-30
fix: Provide explicit template specialisations

phako:
This fixes a huge chunk of the LTO errors we are seeing on noble.

Frankly it is also unclear why it worked in the first place

### DIFF
--- a/include/NumericAddressedBackend.h
+++ b/include/NumericAddressedBackend.h
@@ -158,10 +158,6 @@ namespace ChimeraTK {
     /// mutex for protecting unaligned access
     std::mutex _unalignedAccess;
 
-    template<typename UserType>
-    boost::shared_ptr<NDRegisterAccessor<UserType>> getRegisterAccessor_impl(
-        const RegisterPath& registerPathName, size_t numberOfWords, size_t wordOffsetInRegister, AccessModeFlags flags);
-
     friend NumericAddressedLowLevelTransferElement;
     friend TriggeredPollDistributor;
 
@@ -169,6 +165,10 @@ namespace ChimeraTK {
     friend class NumericAddressedBackendMuxedRegisterAccessor;
 
    private:
+    template<typename UserType>
+    boost::shared_ptr<NDRegisterAccessor<UserType>> getRegisterAccessor_impl(
+        const RegisterPath& registerPathName, size_t numberOfWords, size_t wordOffsetInRegister, AccessModeFlags flags);
+
     // internal helper function to get the a synchronous accessor, which is also needed by the asynchronous version
     // internally, but is not given out
     template<typename UserType>
--- a/include/VirtualFunctionTemplate.h
+++ b/include/VirtualFunctionTemplate.h
@@ -2,14 +2,10 @@
 // SPDX-License-Identifier: LGPL-3.0-or-later
 #pragma once
 
-#include "SupportedUserTypes.h"
-
 #include <boost/bind/bind.hpp>
 #include <boost/function.hpp>
 #include <boost/fusion/include/at_key.hpp>
 #include <boost/fusion/include/for_each.hpp>
-
-#include <tuple>
 
 /** Define a virtual function template with the given function name and
  * signature in the base class. The signature must contain the typename template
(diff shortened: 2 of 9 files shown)


include/async/MuxedInterruptDistributor.h
```diff
@@ -4,6 +4,8 @@
 
 #include "../VersionNumber.h"
 
+#include "VariableDistributor.h"
```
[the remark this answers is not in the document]
phako: This was all removed as it was not necessary

include/NumericAddressedBackend.h
```diff
@@ -158,17 +158,17 @@ namespace ChimeraTK {
...
-        const RegisterPath& registerPathName, size_t numberOfWords, size_t wordOffsetInRegister, AccessModeFlags flags);
-
     friend NumericAddressedLowLevelTransferElement;
     friend TriggeredPollDistributor;
 
     template<class UserType, class ConverterType>
     friend class NumericAddressedBackendMuxedRegisterAccessor;
 
    private:
+    template<typename UserType>
```
killenb, line 168: Wow, yes. Is is now private so the derived class cannot call it and has to use the correct vtable override. I like it!

include/VirtualFunctionTemplate.h
```diff
@@ -30,8 +26,13 @@
  *
  *  In the derived class, the function template must be implemented with the
  * same signature, and the vtable for this virtual function template must be
- * filled using the macros DEFINE_VIRTUAL_FUNCTION_TEMPLATE_VTABLE_FILLER and
- *  FILL_VIRTUAL_FUNCTION_TEMPLATE_VTABLE.
+ * filled differently depending on whether you want to (be able to) call the base class's
+ * implementation or not:
+ *
+ * <ul> If you do not care about the base implementation, use the macros DEFINE_VIRTUAL_FUNCTION_TEMPLATE_VTABLE_FILLER
+ * and FILL_VIRTUAL_FUNCTION_TEMPLATE_VTABLE. <ul> If you want to be able to delegate to the base implementation, use
```
killenb, line 33: It seems the auto code formatter screwed up the listing.

include/VirtualFunctionTemplate.h
```diff
@@ -68,6 +85,19 @@
     pair.second = [this](auto... args) { return this->functionName<VTableFillerUserType>(args...); };                  \
   })
 
+/** Save the old vtable to be accessible by CALL_BASE_FUNCTION_TEMPLATE and overwrite it with the
+ *  new implementation.
+ */
+#define OVERRIDE_VIRTUAL_FUNCTION_TEMPLATE(functionName)                                                               \
+  do {                                                                                                                 \
+    /* Copy the old vtable into _base before filling it with the new implementation */                                 \
+    ChimeraTK::for_each(functionName##_base_vtable.table, [&](auto& pair) {                                            \
```
killenb, line 94: I think this is not going to work in a chain, because the functionName##_base_vtable name is not unique per instance. If ExceptionDummy wants to call the Dummy-Implementation and Dummy wants to call the NumericAddressed implementation, you need three tables.

src/DummyBackendBase.cc
```diff
@@ -37,4 +38,34 @@ namespace ChimeraTK {
     }
   }
 
+  template<typename UserType>
```
killenb, line 41: I did not check that this is the code moved from the header file.

tests/executables_src/testUnifiedTypeChangingDecorator.cpp
```diff
@@ -61,7 +65,8 @@ class DecoratorBackend : public ExceptionDummy {
     auto [path, type] = getPathAndType(registerPathName);
 
     return getTypeChangingDecorator<UserType>(
-        ExceptionDummy::getRegisterAccessor_impl<float>(path, numberOfWords, wordOffsetInRegister, flags), type);
+        CALL_BASE_FUNCTION_TEMPLATE(getRegisterAccessor_impl, float, path, numberOfWords, wordOffsetInRegister, flags),
```
killenb, line 68: What if I want to call NumericAddressedBackend::getRegisterAccessor_impl() or DummyBackend::getRegisterAccessor_impl() here?

include/VirtualFunctionTemplate.h
mhier: The VTABLE should be called identical on all levels (i.e. do not add "_base" to the name in DEFINE_VIRTUAL_FUNCTION_OVERRIDE_VTABLE). This macro should receive a second argument with the name of the direct base class (or maybe auto-detect it with some fancy C++ magic, but it will go wrong in case of multiple inheritance). It can then get the vtable from the specified direct base class.

phako: Fixed

src/DummyBackendBase.cc
killenb: In the diff, this code block is "new", but in fact, it has been moved here from the header. I did not compare the old header and the new .cc file to check that this section is identical in both.

phako: Ah. There is the change to properly chain up of course, but otherwise it should be a move

killenb (approved):
I am impressed how an implementation that is complete nonsense in any real application is ideal and efficient for the test. I like that :-)

mhier (approved):

PR #422, opened 2024-05-02
fix(UIO/XDMA): Backends have to store shared_ptr do Domain

killenb:


mhier:
There is a typo in the commit message, "do" -> "to" I guess?

killenb:
Oh, yes. I will fix that.

phako (approved):
Agree with the code change, but I feel that I am missing the context why this is necessary.

PR #423, opened 2024-05-03
fix(rebot): boost exception escaping when host unreachable

killenb:


killenb (commented):
- testUtilitiies is failing
- coding style test fails

killenb (commented):
Looks good now. You can proceed with the squash merge.
The commit message should be the title of this pull request (I guess this is default for a squash? Have too little experience.)

PR #438, opened 2024-09-25
improvement: acc.read throws if in ReadAnyGroup

d-rothe:
if the accessor is already in a ReadAnyGroup, that must be used for reading

### DIFF
--- a/include/TransferElementAbstractor.h
+++ b/include/TransferElementAbstractor.h
@@ -142,7 +142,7 @@ namespace ChimeraTK {
      *
      * Note: Avoid using this in application code, since it will break the abstraction!
      */
-    const boost::shared_ptr<TransferElement>& getHighLevelImplElement() { return _impl; }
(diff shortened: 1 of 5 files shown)


include/TransferElementAbstractor.h
```diff
@@ -142,7 +142,7 @@ namespace ChimeraTK {
      *
      * Note: Avoid using this in application code, since it will break the abstraction!
      */
-    const boost::shared_ptr<TransferElement>& getHighLevelImplElement() { return _impl; }
+    [[nodiscard]] boost::shared_ptr<TransferElement>& getHighLevelImplElement() { return _impl; }
```
mhier, line 145: This needs to stay const, as we otherwise allow modifying the boost::shared_ptr in the abstractor (e.g. let it point to a different target). That is not what this function shall be used for.

d-rothe: ok, I reverted that change (it was accidental and not required)

mhier (approved):

PR #440, opened 2024-10-08
feat: add type definitions for std::variant with all UserTypes

mhier:


mhier:
This is used by the ApplicationCore Python bindings, but I would like to use it in different context as well and hence move it to DeviceAccess.

killenb (approved):

PR #442, opened 2024-10-10
feat: userTypeToUserType supports HEX strings

fiveisgreen:
userTypeToNumeric_impl<std::string, NUMERIC> was modified to be interpret both decimal and hexidecimal representations of numbers into numeric types. Hexidecimal numbers are identified by 'x' or 'X' as the 2nd character.
This allows userTypeToUserType<UserType,string> to be fed strings representations of hexidecimal numbers and correctly interpret them. 

This was tested for correct behavior of decimal and hexidecimal integers into int, int64, uint, int8_t, uint8_t, and bool.

### DIFF
--- a/include/SupportedUserTypes.h
+++ b/include/SupportedUserTypes.h
@@ -285,26 +285,32 @@ namespace ChimeraTK {
 
   template<typename NUMERIC>
   NUMERIC detail::userTypeToNumeric_impl<std::string, NUMERIC>::impl(const std::string& value) {
-    if constexpr(std::is_same<NUMERIC, Boolean>::value) {
-      // special treatment for Boolean
-      std::stringstream ss(value);
-      Boolean converted;
-      ss >> converted;
-      return converted;
-    }
+    // 0xABCD form hex
+    bool isHexidecimal = (value.length() > 1) and ((value[1] == 'x') or (value[1] == 'X')) and (value[0] == '0'); 
 
     if constexpr(!std::is_same<NUMERIC, int8_t>::value && !std::is_same<NUMERIC, uint8_t>::value) {
       NUMERIC v;
-      std::stringstream ss(value);
+      std::stringstream ss;
+      if(isHexidecimal) {
+        ss << std::hex << value.substr(2);
+      }
+      else {
+        ss.str(value);
+      }
       ss >> v;
       return v;
     }
     else {
       // int8_t and uint8_t are interpreted as char resp. unsigned char, in which case the stream operator does the
       // wrong thing...
-      int temp;
+      int64_t temp;
       try {
-        temp = std::stoi(value);
+        if(isHexidecimal) {
+          temp = std::stoll(value, nullptr, 16);
+        }
+        else {
+          temp = std::stoll(value);
+        }
       }
       catch(...) {
         // ignore any parsing errors and just return 0 in that case
--- a/tests/executables_src/testUtilities.cpp
+++ b/tests/executables_src/testUtilities.cpp
@@ -5,12 +5,14 @@
 #define BOOST_TEST_MODULE DeviceTest
 #include <boost/test/unit_test.hpp>
 using namespace boost::unit_test_framework;
-
 #include "BackendFactory.h"
 #include "DeviceInfoMap.h"
 #include "Exception.h"
 #include "SupportedUserTypes.h"
 #include "Utilities.h"
+
+#include <cstdint>
+#include <limits>
 
 #define VALID_SDM "sdm://./pci:pcieunidummys6;undefined"
 #define VALID_SDM_WITH_PARAMS "sdm://./dummy=goodMapFile.map"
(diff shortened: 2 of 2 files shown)


include/SupportedUserTypes.h
```diff
@@ -285,17 +285,31 @@ namespace ChimeraTK {
 
   template<typename NUMERIC>
   NUMERIC detail::userTypeToNumeric_impl<std::string, NUMERIC>::impl(const std::string& value) {
+    bool isHexidecimal = value.length() > 1 and (value[1] == 'x' or value[1] == 'X'); // OxABCD form hex
```
killenb, line 288: please add parantheses around all logical blocks for clearness. I would add a check that the character before the x is a zero (actually even the comment was wrong, it had a capital "Oh".
```suggestion
    // 0xABCD form hex
    bool isHexidecimal = (value.length() > 1) and ((value[1] == 'x') or (value[1] == 'X')) and (value[0] == '0'); 
```

include/SupportedUserTypes.h
```diff
@@ -285,17 +285,31 @@ namespace ChimeraTK {
...
       Boolean converted;
+      if(isHexidecimal) {
+        ss << std::hex << value;
+      }
+      else {
+        ss.str(value);
+      }
       ss >> converted;
       return converted;
     }
```
killenb, line 302: Actually, I am wondering why there is a special case for Boolean. It is (and was) the same code as for the not  (u)int_8. We should clean this up and delete the unnecessary code for Boolean (or am I overlooking something)?

include/SupportedUserTypes.h
killenb: I put the comment into the line above because the code because the line was getting too long for the coding style.

killenb (changes requested):
In the merge request you write that it has been tested, but I see no tests in the commit.

Please extend the according unit test.

include/SupportedUserTypes.h
fiveisgreen: This special case seems to be safely removable. I tried removing it and was still able to parse into Boolean without a problem.

include/SupportedUserTypes.h
```diff
@@ -285,31 +285,23 @@ namespace ChimeraTK {
...
+    //If value starts with "0x" indicating hexidecimal
+    if(value.length() >= 2 and (value[0] == '0') and ((value[1] == 'x') or (value[1] == 'X'))) {
+        unsigned long long temp;
+        try {
+            temp = std::stoull(value, nullptr, 16);
+        }
+        catch(...) { // ignore any parsing errors and just return 0 in that case
+            temp = 0;
+        }
+        return detail::userTypeToNumeric_impl<unsigned long long, NUMERIC>::impl(temp);
```
killenb, line 298: What happened to the previous implementation with the std::hex in the stringstream? That one was much better because it does not use an intermediate data type.

We have to discuss again what the wanted behaviour is (see review comment of the test file).

tests/executables_src/testUtilities.cpp
```diff
@@ -264,3 +264,58 @@ BOOST_AUTO_TEST_CASE(testUserTypeToUserType_Boolean) {
 }
 
 /**********************************************************************************************************************/
+
+BOOST_AUTO_TEST_CASE(testUserTypeToUserType_String){
+  //Regular decimal branch
+  BOOST_TEST(ChimeraTK::userTypeToUserType<ChimeraTK::Boolean>(std::string("1")) == true);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<ChimeraTK::Boolean>(std::string("2")) == true);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<ChimeraTK::Boolean>(std::string("-123456789")) == true);
```
killenb, line 272: Actually, the whole "to Boolean" part is already covered in the test above. We should not repeat it here but make a code comment where to find it.

tests/executables_src/testUtilities.cpp
```diff
@@ -264,3 +264,58 @@ BOOST_AUTO_TEST_CASE(testUserTypeToUserType_Boolean) {
 }
 
 /**********************************************************************************************************************/
+
+BOOST_AUTO_TEST_CASE(testUserTypeToUserType_String){
+  //Regular decimal branch
```
killenb, line 269: Maybe split the decimal and the hex part into two tests.

tests/executables_src/testUtilities.cpp
```diff
@@ -264,3 +264,58 @@ BOOST_AUTO_TEST_CASE(testUserTypeToUserType_Boolean) {
...
+
+  BOOST_TEST(ChimeraTK::userTypeToUserType<bool>(std::string("0")) == false);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<bool>(std::string("1")) == true);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<bool>(std::string("2")) == true);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<bool>(std::string("-123456789")) == true);//ce
+
+  BOOST_TEST(ChimeraTK::userTypeToUserType<int>(std::string("7777")) == 7777);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<int>(std::string("-7777")) == -7777);
+  BOOST_TEST(ChimeraTK::userTypeToUserType<int32_t>(std::string("-559038737"))==  -559038737);//32bit into int32_t //ce
+  BOOST_TEST(ChimeraTK::userTypeToUserType<int32_t>(std::string("12345678912345"))== 0x7FFFFFFF);//ce
```
killenb, line 282: What is ce?

tests/executables_src/testUtilities.cpp
```diff
@@ -264,3 +264,58 @@ BOOST_AUTO_TEST_CASE(testUserTypeToUserType_Boolean) {
...
+  BOOST_TEST(ChimeraTK::userTypeToUserType<int64_t>(std::string("-12345678912345"))== -12345678912345);
+
+  //int8_t,uint8_t branch
```
killenb, line 285: You are testing code branches, which is an implementation detail. Refactoring could change this, rendering the tests ineffective.

Instead, we should test observable  behaviour:
For each supported user type test (output type):
* Minimum of the dynamic range
* Middle of the dynamic range
* Max of the dynamic range
* Overflow of dynamic range (all directions)
* Invalid input?

 And this kind if test is done for both input types: decimal integers and hexadecimal numbers.
 It would also be needed for decimal floating point and decimal scientific exponential if we want to be complete.

killenb (changes requested):
* The tests should be structured to check observable behaviour, not code branches.
* We have to discuss the wanted behaviour for all user types. I don't have a complete picture yet.

killenb:
Also, don't `merge` the master. `rebase` instead. We in the end want a linear history, and the only solution after a merge is to squash everything into a single commit. This takes away the possibility to have two or more commits with different features in a branch.

killenb (approved):

killenb (approved):
Test is now well structured. Back to an improved version of the original implementation.

PR #447, opened 2024-11-08
Some handy additions triggered by Python binding changes

mhier:


include/NDRegisterAccessorAbstractor.h
```diff
@@ -48,6 +48,10 @@ namespace ChimeraTK {
      */
     const NDRegisterAccessorAbstractor& operator=(const NDRegisterAccessorAbstractor& rightHandSide) const = delete;
 
+    boost::shared_ptr<NDRegisterAccessor<UserType>> getImpl();
+
+    using value_type = UserType;
```
killenb, line 53: value_type is not our coding style. Will the linter complain? What is it used for?

include/ScalarRegisterAccessor.h
killenb, line 1: At the moment, the only valid raw type is int32_t. Is there any place where this could possibly be used?

killenb (approved):

include/ScalarRegisterAccessor.h
mhier: Fundamentally, this is not true. Backends might use any UserType as raw type. Currently there is only NumericAddressedBackends offer raw transfers, and those only use int32_t. We should probably lift that restriction and allow in particular int64_t as well.

I could even imagine string raw data, e.g. in case of the command based backend.

include/NDRegisterAccessorAbstractor.h
mhier: Sorry, this should not have been committed, I removed it.

PR #451, opened 2024-11-26
fix: add missing register name to lmap math plugin exception

mhier:


phako:
Ah sorry, I wanted to click approve -.-

PR #455, opened 2024-12-04
fix(sharedDummy): dead lock due to stale lock

mhier:


backends/SharedDummy/src/SharedMemoryManager.cc
```diff
@@ -5,6 +5,42 @@
...
+    // BOOST versions like on Ubuntu 20.04. This can be removed and replaced with boost::interprocess::named_mutex once
+    // we are fully on Ubuntu 24.04 or newer.
+    class IpcNamedMutex {
+     public:
+      explicit IpcNamedMutex(boost::interprocess::named_mutex& mx) : _mx(mx) {}
+
+      template<typename Duration>
+      // NOLINTNEXTLINE(readability-identifier-naming)
+      bool try_lock_for(const Duration& dur) {
+        auto abs_time = boost::posix_time::microsec_clock::local_time();
```
d-rothe, line 19: this needs to be universal_time()   (I tried it out since docu is not clear)

d-rothe (changes requested):
This does not work cleanly. Tried out following code since I had problems earlier with timed_lock.
Removing and re-creating the mutex in-place does not work for me as expected.
Run following code twice, and you get a failed assertion, showing that the re-created mutex does not work:

#include <boost/date_time.hpp>
#include <iostream>

#include <boost/filesystem.hpp>
#include <boost/interprocess/allocators/allocator.hpp>
#include <boost/interprocess/containers/vector.hpp>
#include <boost/interprocess/managed_shared_memory.hpp>
#include <boost/interprocess/sync/named_mutex.hpp>

using namespace boost::interprocess;

void demo_boost_interprocess(){
    const char* muName = "mu1";
    named_mutex interprocessMutex(open_or_create, muName);

    // is it allowed to destroy a locked interprocess mutex?
    std::unique_lock<boost::interprocess::named_mutex> lock(interprocessMutex);
    // test whether timed_lock works
    //auto abs_time = boost::posix_time::microsec_clock::local_time();
    auto abs_time = boost::posix_time::microsec_clock::universal_time();
    auto bmillis = boost::posix_time::milliseconds(2000);
    abs_time += bmillis;
    std::cout << "try out timed_lock" << std::endl;
    bool ret = interprocessMutex.timed_lock(abs_time);
    assert (ret == false);

    std::cout << "removing locked boost::interprocess::named_mutex returns "  << boost::interprocess::named_mutex::remove(muName) << std::endl;
    interprocessMutex.~named_mutex();
    new(&interprocessMutex) boost::interprocess::named_mutex(boost::interprocess::create_only, muName);

    std::cout << "interprocessMutex created again" << std::endl;
}

backends/SharedDummy/src/SharedMemoryManager.cc
killenb: If I am not mistaken, both universal and local time can have leap seconds and run backwards, so you can  have times "in the past" and the same time twice.
We should use a steady clock. However, I don't know if that's consistent across multiple processes.

mhier: @drothe, I have tested this code to work on both focal and noble. Maybe I did not check everything on focal, but on noble the timeout definitively works as expected, and so does the in-place construction of the mutex. Can you be more specific in which way it does not work for you?

mhier:
> This does not work cleanly. Tried out following code since I had problems earlier with timed_lock. Removing and re-creating the mutex in-place does not work for me as expected. Run following code twice, and you get a failed assertion, showing that the re-created mutex does not work:
> 
> ```
> #include <boost/date_time.hpp>
> #include <iostream>
> 
> #include <boost/filesystem.hpp>
> #include <boost/interprocess/allocators/allocator.hpp>
> #include <boost/interprocess/containers/vector.hpp>
> #include <boost/interprocess/managed_shared_memory.hpp>
> #include <boost/interprocess/sync/named_mutex.hpp>
> 
> using namespace boost::interprocess;
> 
> void demo_boost_interprocess(){
>     const char* muName = "mu1";
>     named_mutex interprocessMutex(open_or_create, muName);
> 
>     // is it allowed to destroy a locked interprocess mutex?
>     std::unique_lock<boost::interprocess::named_mutex> lock(interprocessMutex);
>     // test whether timed_lock works
>     //auto abs_time = boost::posix_time::microsec_clock::local_time();
>     auto abs_time = boost::posix_time::microsec_clock::universal_time();
>     auto bmillis = boost::posix_time::milliseconds(2000);
>     abs_time += bmillis;
>     std::cout << "try out timed_lock" << std::endl;
>     bool ret = interprocessMutex.timed_lock(abs_time);
>     assert (ret == false);
> 
>     std::cout << "removing locked boost::interprocess::named_mutex returns "  << boost::interprocess::named_mutex::remove(muName) << std::endl;
>     interprocessMutex.~named_mutex();
>     new(&interprocessMutex) boost::interprocess::named_mutex(boost::interprocess::create_only, muName);
> 
>     std::cout << "interprocessMutex created again" << std::endl;
> }
> ```

I have tested this code to work on both focal and noble. Maybe I did not check everything on focal, but on noble the timeout definitively works as expected, and so does the in-place construction of the mutex. Can you be more specific in which way it does not work for you?

d-rothe:
> > This does not work cleanly. Tried out following code since I had problems earlier with timed_lock. Removing and re-creating the mutex in-place does not work for me as expected. Run following code twice, and you get a failed assertion, showing that the re-created mutex does not work:
> > ```
> > #include <boost/date_time.hpp>
> > #include <iostream>
> > 
> > #include <boost/filesystem.hpp>
> > #include <boost/interprocess/allocators/allocator.hpp>
> > #include <boost/interprocess/containers/vector.hpp>
> > #include <boost/interprocess/managed_shared_memory.hpp>
> > #include <boost/interprocess/sync/named_mutex.hpp>
> > 
> > using namespace boost::interprocess;
> > 
> > void demo_boost_interprocess(){
> >     const char* muName = "mu1";
> >     named_mutex interprocessMutex(open_or_create, muName);
> > 
> >     // is it allowed to destroy a locked interprocess mutex?
> >     std::unique_lock<boost::interprocess::named_mutex> lock(interprocessMutex);
> >     // test whether timed_lock works
> >     //auto abs_time = boost::posix_time::microsec_clock::local_time();
> >     auto abs_time = boost::posix_time::microsec_clock::universal_time();
> >     auto bmillis = boost::posix_time::milliseconds(2000);
> >     abs_time += bmillis;
> >     std::cout << "try out timed_lock" << std::endl;
> >     bool ret = interprocessMutex.timed_lock(abs_time);
> >     assert (ret == false);
> > 
> >     std::cout << "removing locked boost::interprocess::named_mutex returns "  << boost::interprocess::named_mutex::remove(muName) << std::endl;
> >     interprocessMutex.~named_mutex();
> >     new(&interprocessMutex) boost::interprocess::named_mutex(boost::interprocess::create_only, muName);
> > 
> >     std::cout << "interprocessMutex created again" << std::endl;
> > }
> > ```
> 
> I have tested this code to work on both focal and noble. Maybe I did not check everything on focal, but on noble the timeout definitively works as expected, and so does the in-place construction of the mutex. Can you be more specific in which way it does not work for you?
> 
> EDIT: My assumption is that in case of the problematic dead lock there is only one single process trying to acquire the lock. It might go wrong if this happens concurrently, but I think this is not important. It will be a very rare case.

My idea was to test whether the mutex created again is functional as it should. So my assumption was, that the assertion in the above code should always be true. But if you first clean up /dev/shm/sem.* and then run the above snipplet twice, you get a failed assertion:  
../../demo_boost_interprocess.cpp:25: void demo_boost_interprocess(): Assertion `ret == false' failed.

However, I now understood the problem. The destructor of the unique_lock lock guard "frees" the re-created mutex even though it was not locked, resulting in a lock count=-1. Then, on the next run, you can lock the mutex twice(!), leading to the failed assertion. The problem goes away if you add std::abort() as last line. 
The code clearly shows illegal behavior with a re-created mutex. But your pull-request code should not have this problem, since you start with std::unique_lock(... std::defer_lock). 

So I guess your code is fine once you change to use universal_time()

d-rothe (approved):

PR #458, opened 2024-12-17
fix(xdma): backend must not close itself on error

mhier:
The XDMA backend called internally close() when a recovery of an
exception was attempted. In particular if the recovery was still
unsuccessful, the device was then left in the closed state leading to
logic_error exceptions in subsequent read/write operations. This caused
ApplicationCore applications to terminate immediately after an exception
of an XDMA device was seen.

This is now fixed by using a dedicated flag for the backend opened
status which is independent of the underlying interface status.

### DIFF
--- a/backends/xdma/src/XdmaBackend.cc
+++ b/backends/xdma/src/XdmaBackend.cc
@@ -23,10 +23,11 @@ namespace ChimeraTK {
       if(isFunctional()) {
         return;
       }
-      close();
     }
 
     _ctrlIntf.emplace(_devicePath);
+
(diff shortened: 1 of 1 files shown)


d-rothe (approved):

backends/xdma/src/XdmaBackend.cc
```diff
@@ -3,6 +3,8 @@
 
 #include "XdmaBackend.h"
 
+#include "Utilities.h"
+
```
mhier, line 7: Reviewing the code myself once more shows me that these lines need to be removed...

PR #459, opened 2024-12-19
Drothe/fix xdma exceptions

d-rothe:


mhier:
I changed this to the "draft" status now due to the "comment for discussion" commit, which should not end up in the master branch...

mhier (approved):

PR #460, opened 2025-01-06
chore: Remove broken example plugin

phako:


phako:
I will probably provide a new one as part of the Tango Backend sprint

mhier:
And please don't call it "plugin" then... :-)

killenb (approved):

PR #461, opened 2025-01-15
fix(sharedDummy): catch exception after previous instance interrupted

mhier:
The SharedDummy backend was crashing with a
boost::interprocess::lock_exception when a previous instance was
interrupted in the wrogn place. This is now fixed by catching the
exception and removing the shared memory segment (which is then broken
anyway).

### DIFF
--- a/backends/SharedDummy/include/SharedDummyBackend.h
+++ b/backends/SharedDummy/include/SharedDummyBackend.h
@@ -155,7 +155,7 @@ namespace ChimeraTK {
     }; /* class SharedMemoryManager */
 
     // Managed shared memory object
-    SharedMemoryManager sharedMemoryManager;
+    std::unique_ptr<SharedMemoryManager> sharedMemoryManager;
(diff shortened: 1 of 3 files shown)


d-rothe (approved):
What I don't get is why this exception occurs at all; I would expect that if I interrupt a process owning an ipc lock, that lock stays valid and the next process simply blocks.

mhier:
> What I don't get is why this exception occurs at all; I would expect that if I interrupt a process owning an ipc lock, that lock stays valid and the next process simply blocks.

Good question. The description of the exception states:
```
This is the exception thrown by shared interprocess_mutex family when a deadlock situation is detected or when using a interprocess_condition the interprocess_mutex is not locked 
```
So somehow it is able to detect the deadlock here. I would like a different handling in such case, but since it is an internal lock in the shared memory segment, and not the lock we create on our own, there seems to be not much else what we can do here...

PR #468, opened 2025-02-19
Minor things

mhier:
I recommend review by commit, as they are unrelated.


phako (approved):
Do we need to review the other backends for correct clone() behavior?

PR #475, opened 2025-03-04
fix(lmap): temporary fix for logic_errors in postRead with math plugin

mhier:
This creates rather annoying issues when used with a TransferGroup, in
which case the exception message is totally invisible ("terminate called
without an active exception").

A more permanent solution should execute checks when the formula is
compiled, so the logic_error can be thrown earlier.


killenb (approved):
ATTENTION: you requested to merge into mhier/json-map-file-parser, not into master!

The test is now failing because it is expecting a logic error.

mhier:
> ATTENTION: you requested to merge into mhier/json-map-file-parser, not into master!
> 
> The test is now failing because it is expecting a logic error.

Both fixed. Thanks.

PR #479, opened 2025-03-10
fix: calling activateAsycRead when asyncRead active

killenb:
- fix: async::Domain re-sends values
- fix: LMAP backend re-sends values
- improve the specification
- add unified backend test

### DIFF
--- a/doc/spec_TransferElement.dox
+++ b/doc/spec_TransferElement.dox
@@ -1,7 +1,7 @@
 // put the namespace around the doxygen block so we don't have to give it all the time in the code to get links
 namespace ChimeraTK {
 /**
-\page spec_TransferElement Technical specification: TransferElement V1.1
+\page spec_TransferElement Technical specification: TransferElement V1.2
 
 > **NOTICE FOR FUTURE RELEASES: AVOID CHANGING THE NUMBERING!** The tests refer to the sections, incl. links and unlinked references from tests or other parts of the specification. These break, or even worse become wrong, when they are not changed consistenty!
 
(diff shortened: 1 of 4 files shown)


doc/spec_TransferElement.dox
```diff
@@ -179,9 +179,14 @@ This document describes the behaviour of the TransferElement base class, the NDR
...
       - \anchor transferElement_B_8_5_3 8.5.3 If a transfer element is created while the device is opened, functional and  Device::activateAsyncRead() has been called, the asynchronous read is activated immediately (incl. sending of the initial value). This ensures that either all or none of the transfer elements are receiving asyncronously send data. [\ref UnifiedTest_TransferElement_B_8_5_3 "Needs to be changed, most probably testing the wrong thing. U"]
-      - 8.5.4 If Device::activateAsyncRead() is called while the device is not opened or has an error, this call has no effect. If it is called when no deactivated transfer element exists, this call also has no effect.
+      - \anchor transferElement_B_8_5_4 8.5.4 A Device::activateAsyncRead() call has no effect if
+         - 8.5.4.1 the device is not opened
+         - 8.5.4.2 the device has an error
+         - 8.5.4.3 no transfer elements without wait_for_new_data exist
```
killenb, line 185: This point is wrong. It should have read
```no transfer elements with wait_for_new_data exist```, but that is also wrong, already according to this spec.
The minimum change is: The backend internally stores that activateAsyncRead has been called to fulfill B.8.5.3 when a new accessor with the flag is created, even if there are no such accessors yet. As all backends are in principle prepared for async accessors (the async domains container is already in the base classes) this conceptually always the case.

File: doc/spec_TransferElement.dox [hunk 2]
```diff
@@ -179,9 +179,14 @@ This document describes the behaviour of the TransferElement base class, the NDR
...
+         - \anchor transferElement_B_8_5_4_4 8.5.4.4 asynchronous read is already active [\ref UnifiedTest_TransferElement_B_8_5_4_4 "U"]
       - 8.5.5 Device::activateAsyncRead() does not throw any exceptions.
       - 8.5.6 When Device::activateAsyncRead() returns, it is not guaranteed that all initial values have been received already.
+      - 8.5.7 Device::activateAsyncRead() is thread safe against other calls of activateAsyncRead(), the creation/destuction of accessors, and read/write operations
```
killenb, line 189: propagation of exceptions is missing.

killenb: done in fixup

phako (approved):

vaeng (approved):

PR #480, opened 2025-03-11
fix: build with newer boost versions

mhier:


mhier:
This fixes compilation on Tumbleweed

backends/Rebot/src/Connection.h
```diff
@@ -40,7 +40,7 @@ namespace ChimeraTK::Rebot {
    private:
     std::string address_;
     const std::string port_;
-    boost::asio::io_service ioService_;
+    boost::asio::io_context ioService_;
```
killenb, line 43: We should also rename the local variables xxxService into xxxContext.

killenb (approved):
Approved after changing the local variable names and if it compiles on bookworm as well.

backends/Rebot/src/Connection.h
mhier: done

phako:
Don't we have a ticket for this somewhere?

killenb:
> Don't we have a ticket for this somewhere?

I don't think so.

PR #482, opened 2025-03-27
Drothe/extended data matching

d-rothe:
the 4 commits can be reviewed independently

### DIFF
--- a/include/DataConsistencyGroup.h
+++ b/include/DataConsistencyGroup.h
@@ -3,108 +3,113 @@
 #pragma once
 
 #include "TransferElementAbstractor.h"
-#include "VersionNumber.h"
-
-#include <unordered_set>
 
 namespace ChimeraTK {
 
   /********************************************************************************************************************/
 
+  namespace DataConsistencyGroupDetail {
+    class MatcherBase;
+  }
+
   /**
    * Group several registers (= TransferElement) which ensures data consistency across multiple variables through an
    * algorithm which matches the VersionNumber. This group does not read on its own. It should work together with a
-   * ReadAnyGroup. You should wait for changed variable and transfer it to this group by calling
-   * ChimeraTK::DataConsistencyGroup::update. If a consistent state is reached, this function returns true.
+   * ReadAnyGroup.
+   * This is a proxy class, which delegates to two different implementations.
+   * MatchingMode=exact is handled by SimpleMatcher (the legacy DataConsistencyGroup implementation).
+   * For this, you should wait for changed variable and transfer it to this group by calling
+   * DataConsistencyGroup::update. If a consistent state is reached, this function returns true.
+   * MatchingMode=historized is handled by HistorizedMatcher. In this case, the provided accessors are decorated
+   * with DataConsistencyDecorators, with the effect that the readAny() returns only on consistent inputs.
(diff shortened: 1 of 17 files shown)


include/ReadAnyGroup.h
```diff
@@ -155,7 +155,8 @@ namespace ChimeraTK {
     void add(TransferElementAbstractor& element);
 
     ReadAnyGroup(const ReadAnyGroup&) = delete;
-    ReadAnyGroup(ReadAnyGroup&& other) noexcept;
+    ReadAnyGroup(ReadAnyGroup&& other) noexcept { operator=(std::move(other)); }
+    ReadAnyGroup& operator=(ReadAnyGroup&& other) noexcept;
```
mhier, line 159: The commit "fixup: add ReadAnyGroup move assignment" probably should be squashed, too? (slightly misformatted, so autosquash wouldn't get it)

include/DataConsistencyGroup.h
```diff
@@ -3,107 +3,111 @@
...
 
 #include "TransferElementAbstractor.h"
-#include "VersionNumber.h"
-#include <unordered_set>
 
 namespace ChimeraTK {
 
   /********************************************************************************************************************/
 
+  class BaseMatcher;
```
mhier, line 11: Better place this and the two implementations into their own namespace, e.g. ChimeraTK::DataConsistencyGroupDetail. Also it should be `MatcherBase`, since it is the base class for the matchers, not a matcher for bases...

include/HistorizedMatcher.h
```diff
@@ -0,0 +1,129 @@
+// SPDX-FileCopyrightText: Deutsches Elektronen-Synchrotron DESY, MSK, ChimeraTK Project <chimeratk-support@desy.de>
```
mhier, line 1: Either place the file into a subdirectory, or change its name to something like `DataConsistencyGroupHistorizedMatcher`. Yes it's long, but nobody is using it directly.

include/SimpleMatcher.h
```diff
@@ -0,0 +1,61 @@
+// SPDX-FileCopyrightText: Deutsches Elektronen-Synchrotron DESY, MSK, ChimeraTK Project <chimeratk-support@desy.de>
```
mhier, line 1: dito for the file name...

src/DataConsistencyDecorator.cc
```diff
@@ -0,0 +1,98 @@
...
+    // target<->h1, so target oldest data (h2) which may be overwritten next. In match search (after
+    // postRead), the three buffers {target, h1, h2} should be searched.
+    _hGroup->updateHistory(_target->getId());
+
+    // Differently from usual decorator behavior, we call target->postRead already here,
+    // because we need user buffer content to judge data consistency.
+    _target->postRead(TransferType::read, true);
+    std::cout << "readCallback: seeing update for target " << _target->getName() << " vs "
+              << _target->getVersionNumber() << std::endl;
+
```
mhier, line 82: remove debug couts (there are more!)

mhier (approved):

include/ReadAnyGroup.h
d-rothe: yes. will force push squashed version now.

PR #486, opened 2025-04-08
feat: add user type as value type

vaeng:


mhier (approved):
Approved - assuming it actually helps with your problem (please check before merging).

PR #498, opened 2025-05-30
Fix/compiler warnings

mhier:
@killenb Is there a reason why we should not yet merge this?


killenb (approved):

killenb:
This was just forgotten. Should be merged.

PR #500, opened 2025-06-05
fix: shm dummy instance id hash

d-rothe:
generate hash from address and all cdd parameters, even when unused. Like this, different cdds result in different shm instances. Put shm segment name creation to Utilities so its available for test code.

NOTE: This change breaks compatibility for shm dummy communication with previous DeviceAccess versions.

### DIFF
--- a/backends/SharedDummy/src/SharedDummyBackend.cc
+++ b/backends/SharedDummy/src/SharedDummyBackend.cc
@@ -7,6 +7,7 @@
 #include "Exception.h"
 #include "MapFileParser.h"
 #include "parserUtilities.h"
+#include "Utilities.h"
 
 #include <boost/filesystem.hpp>
 #include <boost/lambda/lambda.hpp>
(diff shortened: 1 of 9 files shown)


backends/SharedDummy/src/SharedDummyBackend.cc
```diff
@@ -130,6 +128,11 @@ namespace ChimeraTK {
 
   boost::shared_ptr<DeviceBackend> SharedDummyBackend::createInstance(
       std::string address, std::map<std::string, std::string> parameters) {
+    // create instanceId from address and parameters.
+    // note, this approach is not perfect: in case two different device URIs are created by reordering their parameters
+    // and URIs are used from same process, it will create two SharedDummyBackend instances (since URI is key in
+    // existingBackends map of factory) although it shouldn't!
```
mhier, line 134: I think this comment should be a bit more elaborate, e.g.:

```suggestion
    // note, this approach is not perfect: in case two different device URIs are created by reordering their parameters
    // and URIs are used from same process, the BackendFactory will create two SharedDummyBackend instances (since it uses
    // the URI string as a key). The order of parameters is lost in the parameters map, so we will here assume the same
    // instanceIdHash. That will lead further down to some error, since we will try to access the same shared memory segment
    // twice in the same process.
```

I hope this summarises your findings properly?

src/Utilities.cpp
```diff
@@ -348,6 +349,27 @@ namespace ChimeraTK {
     }
   }
 
+  std::size_t Utilities::shmDummyInstanceIdHash(
```
mhier, line 352:
```suggestion
/********************************************************************************************************************/

  std::size_t Utilities::shmDummyInstanceIdHash(
```

src/Utilities.cpp
```diff
@@ -348,6 +349,27 @@ namespace ChimeraTK {
...
+      const std::string& address, const std::map<std::string, std::string>& parameters) {
+    // To find hash, concatenate in ordered way, address and parameters
+    std::string str = address;
+    for(const auto& e : parameters) {
+      str += e.first + "=" + e.second + ";";
+    }
+    return std::hash<std::string>{}(str);
+  }
+
+  std::string Utilities::createShmName(
```
mhier, line 362:
```suggestion
/********************************************************************************************************************/

  std::string Utilities::createShmName(
```

mhier (approved):
Looks great, besides the functional change, it is also a good structural improvement. I suggest only minor comment changes.

PR #502, opened 2025-06-20
feat(lmap): Convenience plugins for system tags

phako:
Add isStatusOutput and hasReverseRecovery plugins as a more condensed
way to add well-known ChimeraTK tags to registers


phako:
As discussed during the focus, but omitted because of time

PR #511, opened 2025-08-29
Test/b.4.2.5

mhier:
See: https://redmine.msktools.desy.de/issues/14286

### DIFF
--- a/backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
+++ b/backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
@@ -186,7 +186,11 @@ namespace ChimeraTK::LNMBackend {
       this->_dataValidity = _secondBufferReg->dataValidity();
     }
 
-    if(!hasNewData) return;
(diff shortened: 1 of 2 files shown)


backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
```diff
@@ -186,7 +186,11 @@ namespace ChimeraTK::LNMBackend {
       this->_dataValidity = _secondBufferReg->dataValidity();
     }
 
-    if(!hasNewData) return;
+    // Note: TransferElement Spec E.6.1 dictates that the version number and data validity needs to be set before this
```
phako, line 189: Isn't that just a re-format?

phako (approved):

backends/LogicalNameMapping/src/LNMDoubleBufferPlugin.cc
mhier: fixed

PR #517, opened 2025-09-22
fix: userTypeToUserType with 0x

fiveisgreen:
In userTypeToNumeric_impl<std::string, NUMERIC>::impl, NUMERIC v had been left uninitialized. This relied on v being initialized to 0, which doesn't work on Bookworm-release, causing the function to fail. Now it's explicitly initialized to 0.
This should solve error seen on [Jenkins](https://jenkins-sw.msktools.desy.de/job/Dragon%20Nightly%20Reporters/job/ChimeraTK-DeviceAccess-CommandBasedBackend/405/testReport/projectroot/tests/Generate_report___bookworm_release___bookworm_release_testStringUtils/).

### DIFF
--- a/include/SupportedUserTypes.h
+++ b/include/SupportedUserTypes.h
@@ -303,7 +303,7 @@ namespace ChimeraTK {
     bool isHexidecimal = (value.length() > 1) and ((value[1] == 'x') or (value[1] == 'X')) and (value[0] == '0');
 
     if constexpr(!std::is_same<NUMERIC, int8_t>::value && !std::is_same<NUMERIC, uint8_t>::value) {
-      NUMERIC v;
+      NUMERIC v = 0;
       std::stringstream ss;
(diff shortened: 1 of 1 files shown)


fiveisgreen:
When the input is "0x", empty string is piped into the stringstream, leaving its buffer empty. Then the `>>` operator tries to parse the buffer which it finds to be empty (eof situation), resulting in a failed parse (the  stringstream failbit and eofbits get raised) and v is not assigned at all by the `>>` operator.

fiveisgreen:
An alternative would be checking `if( ss.fail() )`, then assigning 0.

killenb (approved):
I am still wondering what the wanted behaviour in case of "0x" is. It is not a valid input.

Before the behaviour was undefined (as seen in different results for release and debug builds, and across distributions). Not it consistently returns 0.

PR #519, opened 2025-09-24
Feat 64 bit map file parser ns

nshehz:


src/TraditionalMapFileParser.cc
```diff
@@ -361,7 +361,7 @@ namespace ChimeraTK::detail {
       channels.emplace_back(NumericAddressedRegisterInfo::ChannelInfo{uint32_t(channel.address - pl.address) * 8,
           channel.type, channel.width, channel.nFractionalBits, channel.signedFlag});
       bytesPerBlock += channel.nBytes;
-      if(channel.nBytes != 1 && channel.nBytes != 2 && channel.nBytes != 4) {
+      if(channel.nBytes != 1 && channel.nBytes != 2 && channel.nBytes != 4 && channel.nBytes != 8) {
```
killenb, line 364: Just a comment. This is the proper extension for the check for primitive types, so the change is fine like this. However, I don't know if this is a necessary restriction. Why can the alignment not be  3 bytes?

tests/IncorrectRegisterWidth.map
```diff
@@ -1,2 +1,2 @@
 # name                    number of elements       address          size           bar    width   fracbits    signed   
-WORD_FIRMWARE                     0x00000001    0x00000000    0x00000004    0x00000000       33         0         0
+WORD_FIRMWARE                     0x00000001    0x00000000    0x00000004    0x00000000       65         0         0
```
killenb, line 2: Ah, and there's the catch. The old test (4 bytes size and 33 bits width) should fail as well. With the changes implementation it does not.
What also should fail is 2 elements, 4 bytes, 32 bits width, while 1 element, 4 bytes, 32 bits is fine.

tests/executables_src/testMapFileParser.cpp
```diff
@@ -139,7 +138,7 @@ BOOST_AUTO_TEST_CASE(testGoodMappFileParse) {
   ChimeraTK::MapFileParser map_file_parser;
   auto [regcat, mdcat] = map_file_parser.parse("goodMapFile.map");
 
-  BOOST_CHECK_EQUAL(regcat.getNumberOfRegisters(), 22);
+  BOOST_CHECK_EQUAL(regcat.getNumberOfRegisters(), 23);
```
killenb, line 141: I was wondering if this is a useful test at all. It always fails when you modify the goodMapFile. The test is useful because getNumberOfRegisters() is public API. But we might want a dedicated test, not run it parasitically on whatever else the "goodMapFile" is testing.

tests/goodMapFile.map
```diff
@@ -39,3 +39,4 @@ LARGE_BAR.NUMBER                  0x00000001    0x00000000    0x00000004    0x10
 MODULE0.INTERRUPT_VOID1               0x00000000    0x00000000    0x00000000    0x00000000       0         0        0 INTERRUPT3 
 MODULE0.INTERRUPT_VOID2               0             0             0             0                0         0        0 INTERRUPT2 
 MODULE0.INTERRUPT_TYPE                0x00000001    0x00000068    0x00000004    0x00000001       18        5        0 INTERRUPT6  
+MODULE1.TEST_AREA2                    0x00000001    0x00000070    0x00000008    0x00000001       32        0        0
```
killenb, line 42: This is not an area. "area" was the old name of 1D registers, or memory blocks with multiple scalars that for technical reasons are somehow connected (e.g. memory region mapped from a sub device into the PCIe address space).

This register is a 64 bit scalar. Also. don't put it into the interrupt section. It should be higher up in the file.

killenb (changes requested):

tests/IncorrectRegisterWidth.map
nshehz: So, we keep it as it is?

src/RegisterCatalogue.cc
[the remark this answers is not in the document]
nshehz: I already had pushed my branch IIRC. I pulled master then rebased master in my branch then pushed branch again.

tests/IncorrectRegisterWidth.map
killenb: I guess we need a few more test cases. The old one is one of them.

tests/newSequences.mapp
```diff
@@ -59,9 +59,9 @@ TEST.MEM_MULTIPLEXED_DMA.SUPER_5 1 0x1E 2 0xD 16 -2 1
 
 # mixed word sizes
 TEST.MEM_MULTIPLEXED_MIXED 6 0 24 0x3
-TEST.MEM_MULTIPLEXED_MIXED.0 1 0x00 1 0x3
-TEST.MEM_MULTIPLEXED_MIXED.1 1 0x01 2 0x3
-TEST.MEM_MULTIPLEXED_MIXED.2 1 0x03 4 0x3
+TEST.MEM_MULTIPLEXED_MIXED.0 1 0x00 1 0x3  8
+TEST.MEM_MULTIPLEXED_MIXED.1 1 0x01 2 0x3  16
+TEST.MEM_MULTIPLEXED_MIXED.2 1 0x03 4 0x3  32
```
killenb, line 64:
```suggestion
TEST.MEM_MULTIPLEXED_MIXED.2 1 0x03 4 0x3  32
TEST.MEM_MULTIPLEXED_MIXED.3 1 0x07 8 0x3  64
```
This one should be extended with 64 bits

tests/sequences.map
```diff
@@ -58,9 +58,9 @@ TEST.SEQUENCE_DMA_15 1 0x1E 2 0xD 16 -2 1
 
 # mixed word sizes
 TEST.AREA_MULTIPLEXED_SEQUENCE_MIXED 6 0 24 0x3
```
killenb, line 60:
```suggestion
TEST.AREA_MULTIPLEXED_SEQUENCE_MIXED 6 0 48 0x3
```

tests/sequences.map
```diff
@@ -58,9 +58,9 @@ TEST.SEQUENCE_DMA_15 1 0x1E 2 0xD 16 -2 1
...
-TEST.SEQUENCE_MIXED_0 1 0x00 1 0x3
-TEST.SEQUENCE_MIXED_1 1 0x01 2 0x3
-TEST.SEQUENCE_MIXED_2 1 0x03 4 0x3
+TEST.SEQUENCE_MIXED_0 1 0x00 1 0x3  8
+TEST.SEQUENCE_MIXED_1 1 0x01 2 0x3  16
+TEST.SEQUENCE_MIXED_2 1 0x03 4 0x3  32
```
killenb, line 63:
```suggestion
TEST.SEQUENCE_MIXED_2 1 0x03 4 0x3  32
TEST.SEQUENCE_MIXED_3 1 0x07 8 0x3  64
```

src/TraditionalMapFileParser.cc
```diff
@@ -385,16 +383,22 @@ namespace ChimeraTK::detail {
...
     auto registerInfo = NumericAddressedRegisterInfo(
         name2D, pl.bar, pl.address, nBlocks, bytesPerBlock * 8, channels, pl.registerAccess, pl.interruptID);
     _pmap.addRegister(registerInfo);
 
     // create 1D entry for reading the multiplexed raw data
+
     assert(pl.nBytes % 4 == 0);
+    auto regSize = 4;
+    if(pl.width == 64) {
+      regSize = 8;
```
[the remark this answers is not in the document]
killenb: I was thinking about that, too. There is no valid use case for it, because how many channels do you want to put into 0 bytes? And why do you want to multiplex 0 channels with 0 entries each?

However, it would be nice not to crash with a "divide by zero" in case of invalid map files. We should throw a ctk::logic_error if nBytes is 0 in this case.

tests/goodMapFile_withoutModules.map
```diff
@@ -8,7 +8,7 @@ WORD_STATUS                       0x00000001    0x00000008    0x00000004    0x00
 WORD_USER1                        0x00000001    0x0000000C    0x00000004    0x00000000       32         0         
 WORD_USER2                        0x00000001    0x00000010    0x00000004    0x00000000       32         0         0
 
-AREA_MULTIPLEXED_SEQUENCE_INT2D           15             0            60             0       16         0         0
+AREA_MULTIPLEXED_SEQUENCE_INT2D           15             0            60             0       32         0         0
```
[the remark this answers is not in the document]
nshehz: corrected.

src/TraditionalMapFileParser.cc
```diff
@@ -200,22 +200,20 @@ namespace ChimeraTK::detail {
...
     // Note: default values for optional information are set in ParsedLine declaration
-
     // extract bar
     is >> std::setbase(0) >> pl.bar;
 
     // extract width
     if(!is.fail()) {
       is >> std::setbase(0) >> pl.width;
-      if(pl.width > 64) {
+      if(pl.nElements > 0 && pl.width > pl.nBytes * 8 / pl.nElements) {
```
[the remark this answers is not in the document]
nshehz: done

PR #520, opened 2025-09-25
docs: Updating dependencies in README.md

cweickhmann:
Though the actual list might change in the future, I'd find it helpful to briefly point out the other dependencies.


killenb (changes requested):
Thanks for your contribution and welcome to ChimeraTK!

Info about our workflow:
We use conventional commits https://www.conventionalcommits.org/en/v1.0.0/ for all of the commits and do not squash branches during merging. All of the commits will stay as they are.

I have seen that the merge request has the "docs:" prefix. However, in our workflow this is not used. Can you please either prefix the individual commits or squash already on the branch?

Thanks

PR #526, opened 2025-10-13
feat: Allow different RawData types

toseko:


include/FixedPointConverter.h
```diff
@@ -331,87 +334,108 @@ namespace ChimeraTK {
...
+      // since this is our raw type. The conversion will properly round, when
+      // needed. Negative overflow exceptions need to be caught for some corner
+      // cases (e.g. number of fractional bits >= number of bits in total).
+      // Positive overflow cannot happen due to the range check above (the negative
+      // branch has one more possible value).
+      RawType raw;
+      try {
+        if(_isSigned) {
+          using converter_signed = boost::numeric::converter<int32_t, double,
+              boost::numeric::conversion_traits<int32_t, double>, boost::numeric::def_overflow_handler, Round<double>>;
```
killenb: This one has to convert to, argh., the signed type with the same number of bits as the RawType.

include/FixedPointConverter.h
```diff
@@ -444,13 +468,83 @@ namespace ChimeraTK {
...
+      _signBitMask = (_isSigned ? 1 << (nBits - 1) : 0x0); // the highest valid bit is the sign
+    }
+    else {
+      _signBitMask = 0;
+    }
+
+    // compute masks of used and unused bits
+    // by using 1L (64 bit) this also works for 32 bits, which needs 33 bits during the calculation
+    // NOLINTNEXTLINE(hicpp-signed-bitwise)
+    _usedBitsMask = static_cast<RawType>((1L << nBits) - 1L);
```
killenb: This hack is gonna fail for 64 bit integers because we would need 65 a bit container.

include/FixedPointConverter.h
```diff
@@ -444,13 +468,83 @@ namespace ChimeraTK {
...
+                                                 // possible value
+                                                 // (0 if unsigned)
+
+    // fill all user type depending values: min and max cooked values and
+    // fractional bit coefficients note: we loop over one of the maps only, but
+    // initCoefficients() will fill all maps!
+    boost::fusion::for_each(_minCookedValues, initCoefficients(this));
+  }
 
+  using DEPRECATED_FIXEDPOINT_DEFAULT = int64_t;
```
killenb: Nopw, that would be int32_t

include/FixedPointConverter.h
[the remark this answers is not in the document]
toseko: Yes, I agree. I think I even saw it, but somehow did not change.

include/FixedPointConverter.h
```diff
@@ -444,13 +468,83 @@ namespace ChimeraTK {
   };
 
   /********************************************************************************************************************/
-  template<>
-  [[nodiscard]] uint32_t FixedPointConverter::toRaw<std::string>(std::string cookedValue) const;
+  /* As the class is templated now, move implementation from the source file*/
+  /********************************************************************************************************************/
+  template<typename RawType>
+  const int FixedPointConverter<RawType>::zero = 0;
+
```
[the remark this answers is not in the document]
toseko: OK.

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
```diff
@@ -159,7 +159,7 @@ namespace ChimeraTK::LNMBackend {
             "Register \"" + TransferElement::getName() + "\" with BitRange plugin is not writeable.");
       }
 
-      auto value = fixedPointConverter.toRaw(buffer_2D[0][0]);
+      auto value = (uint32_t)fixedPointConverter.toRaw(buffer_2D[0][0]);
```
killenb: dito

killenb (commented):
maybe we should make a static assert for the moment that the FixedPointConverter is only instantiated with int32_t?

Apart from that I think we are ready to merge to the master. I would do a full final review af the remaining issues are fixed (only did the delta of the comments and the commit, and lost the overview)

include/FixedPointConverter.h
killenb: Yep, that looks good.

include/FixedPointConverter.h
```diff
@@ -241,81 +242,88 @@ namespace ChimeraTK {
...
       COOKED_ITERATOR cooked_begin) {
     // Handle integer and floating-point types differently.
     switch(boost::fusion::at_key<UserType>(fpc.conversionBranch_toCooked.table)) {
       case 1: { // std::numeric_limits<UserType>::is_integer && _fpc->_fractionalBits == 0 && !_fpc->_isSigned
-        std::transform(raw_begin, raw_end, cooked_begin, [&fpc](int32_t rawValue) {
+        std::transform(raw_begin, raw_end, cooked_begin, [&fpc](RawType rawValue) {
           fpc.padUnusedBits(rawValue);
-          return numericToUserType<UserType>(*(reinterpret_cast<uint32_t*>(&rawValue)));
+          if constexpr(sizeof(RawType) == 4) {
+            return numericToUserType<UserType>(*(std::bit_cast<uint32_t*>(&rawValue)));
```
killenb: ```std::bit_cast``` works on an object, not on a pointer (in contrast to the reinterpret_cast), at least according to the docu. I don't know if this does the right thing. Probably it does, because it changes one pointer type into another.

At least it's unnecessarily complicated. 
```
numericToUserType<UserType>(std::bit_cast<uint32_t>(rawValue));
```
should do the trick.

include/FixedPointConverter.h
```diff
@@ -241,81 +242,88 @@ namespace ChimeraTK {
...
+          }
+          else if constexpr(sizeof(RawType) == 8) {
```
killenb: I am wondering if we should separate the 64 bit case into another commit. 

1. commit: Formally introduce RawType as template parameter, but have an assertion in the constructor that only allows int32_t as type.
2. commit: Test extension to 64 bits, then make it work.

We first get commit 1. ready and put it into the master before we are messing with too may things.

tests/executables_src/testFixedPointConverter.cpp
```diff
@@ -25,6 +25,9 @@ using namespace boost::unit_test_framework;
 #define SIGNED_HEX_TO_DOUBLE(INPUT) static_cast<double>(static_cast<int32_t>(INPUT))
 #define SIGNED_HEX_TO_INT64(INPUT) static_cast<int64_t>(static_cast<int32_t>(INPUT))
 
+#define SIGNED_HEX16_TO_DOUBLE(INPUT) static_cast<double>(static_cast<int16_t>(INPUT))
```
killenb: What about SIGNED_HEX64_TOxxx ?

include/FixedPointConverter.h
```diff
@@ -399,14 +406,32 @@ namespace ChimeraTK {
       RawType raw;
       try {
         if(_isSigned) {
-          using converter_signed = boost::numeric::converter<int32_t, double,
-              boost::numeric::conversion_traits<int32_t, double>, boost::numeric::def_overflow_handler, Round<double>>;
-          raw = converter_signed::convert(d_cooked);
+          if constexpr(sizeof(RawType) == 4) {
```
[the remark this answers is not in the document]
killenb: We do not touch the code structure for now.

tests/executables_src/testFixedPointConverter.cpp
```diff
@@ -904,34 +922,74 @@ BOOST_AUTO_TEST_CASE(testInt32_fraction32) {
...
-  checkToRaw(converter, -5.75, 0x80000000);
+  checkToRaw(converter, 0.75, 0x7FFFFFFF, "ToRaw3");
+  checkToRaw(converter, -0.75, 0x80000000, "ToRaw4");
+  checkToRaw(converter, 3.25, 0x7FFFFFFF, "ToRaw5");
+  checkToRaw(converter, -3.25, 0x80000000, "ToRaw6");
+  checkToRaw(converter, 5.75, 0x7FFFFFFF, "ToRaw7");
+  checkToRaw(converter, -5.75, 0x80000000, "ToRaw8");
 
   checkToCooked(converter, 0x40000000, 0.25);
   checkToCooked(converter, 0xC0000000, -0.25);
```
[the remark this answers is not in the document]
killenb: Code seems to have resolved whatever this comment originally referred to.

include/FixedPointConverter.h
killenb: I will move this to a different branch and directly resolve it based on the master

toseko:
This is old pull request. Part it consider was rewritten from scratch so it become obsolete.

PR #565, opened 2026-06-08
fix: Use target's device for sharedAccessorMap key

phako:
Previously using {lmap backend, target device name} as key, which would lead to shared access
to random accessors if the target device name was identical.

Now using {target device backend, target device name} as key for the sharedAccessorMap.

Otherwise one would unwillingly share random registers that happen to be
named similarly.

Same implementation as used in the BitAccesor

### DIFF
--- a/backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
+++ b/backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
@@ -82,11 +82,18 @@ namespace ChimeraTK::LNMBackend {
 
     /******************************************************************************************************************/
     BitRangeAccessPluginDecorator(boost::shared_ptr<LogicalNameMappingBackend>& backend,
+        boost::shared_ptr<DeviceBackend>& targetDevice,
         const boost::shared_ptr<ChimeraTK::NDRegisterAccessor<uint64_t>>& target, const std::string& name,
         uint64_t shift, uint64_t numberOfBits, uint64_t dataInterpretationFractionalBits,
         uint64_t dataInterpretationIsSigned)
     : ChimeraTK::NDRegisterAccessorDecorator<UserType, uint64_t>(target), _shift(shift), _numberOfBits(numberOfBits),
       _writeable{_target->isWriteable()} {
+      // Reset the version number. The target accessor may be shared between different decorators (e.g. multiple
+      // bit-range registers targeting the same physical register). In that case the target's version number may have
+      // been set by operations through another decorator, but from the user's perspective this is a fresh accessor.
+      // The test UnifiedBackendTest_B_6 checks this by verifying VersionNumber(nullptr).
+      this->_versionNumber = VersionNumber{nullptr};
+
       // makeConverterLoopHelper expects a NumericAddressedBackend RegisterInfo, which we create with the relevant
       // parameters.
       NumericAddressedRegisterInfo registerInfo{name, 1, 0, sizeof(uint64_t), 0, uint32_t(numberOfBits),
@@ -100,9 +107,10 @@ namespace ChimeraTK::LNMBackend {
       }
 
       auto& map = boost::fusion::at_key<uint64_t>(backend->sharedAccessorMap.table);
+
       RegisterPath path{name};
       path.setAltSeparator(".");
-      LogicalNameMappingBackend::AccessorKey key{backend.get(), path};
+      LogicalNameMappingBackend::AccessorKey key{targetDevice.get(), path};
 
       auto it = map.find(key);
       if(it != map.end()) {
@@ -194,7 +202,7 @@ namespace ChimeraTK::LNMBackend {
 
         // FIXME: Not setting the data validity according to the spec point B2.5.1.
         // This needs a change in the fixedpoint converter to tell us that it has clamped the value to reliably work.
-        // To be revisted after fixing https://redmine.msktools.desy.de/issues/12912
+        // To be revisited after fixing https://redmine.msktools.desy.de/issues/12912
 
         // When in a transfer group, only the first accessor to write to the _target can call read() in its preWrite()
         // Otherwise it will overwrite the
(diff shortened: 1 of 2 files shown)


mhier:
shouldn't you then also use the target register name? Imagine two logical name mapping instances pointing to the same target device, both having an identically name redirectedBit (or whatever) but pointing to different registers in the target...

phako:
> shouldn't you then also use the target register name? Imagine two logical name mapping instances pointing to the same target device, both having an identically name redirectedBit (or whatever) but pointing to different registers in the target...

Sorry, that is probably unclear from the description. It now uses {target device backend, target name], previously it used {lmap backend, target name} as key.

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
```diff
@@ -328,8 +330,17 @@ namespace ChimeraTK::LNMBackend {
       boost::shared_ptr<LogicalNameMappingBackend>& backend, boost::shared_ptr<NDRegisterAccessor<TargetType>>& target,
       const UndecoratedParams& params) {
     if constexpr(std::is_same_v<TargetType, uint64_t>) {
-      return boost::make_shared<BitRangeAccessPluginDecorator<UserType>>(backend, target, params._name, _shift,
-          _numberOfBits, dataInterpretationFractionalBits, dataInterpretationIsSigned);
+      // Resolve the correct target device (must match the logic in getAccessor_impl)
```
killenb, line 353: I am not a fan of implementing the same logic at two places, and it has to consistently be the same to work. However, I don't see a way around it with the current design of "decorateAccessor()", which has to know the constructor signature.

This is just a comment. You can resolve it when read. If you have an idea how to improve it you can create a ticket. Otherwise just proceed with the merge.

killenb (approved):

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
phako: One could at least refactor the lookup pattern into a method of the LNM backend.

phako:
Actually, this breaks the Unified Backend Test

killenb:
> Actually, this breaks the Unified Backend Test

How? Why? It should not. Maybe the test relies on the behaviour being wrong?

phako:
```
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1711): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1719): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed
    -> runtime_error case: 0
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1735): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed
    registerName = /BitRangeMiddle
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1711): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1719): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed
    -> runtime_error case: 0
/home/jgeorg/Source/workflow-scripts/sources/ChimeraTK-DeviceAccess/./include/UnifiedBackendTest.h(1735): error: in "LMapBackendUnifiedTestSuite/unifiedBackendTest": check reg.getVersionNumber() == VersionNumber(nullptr) has failed

```

mhier:
> > shouldn't you then also use the target register name? Imagine two logical name mapping instances pointing to the same target device, both having an identically name redirectedBit (or whatever) but pointing to different registers in the target...
> 
> Sorry, that is probably unclear from the description. It now uses {target device backend, target name], previously it used {lmap backend, target name} as key.

Ah, `_info.registerName` is the target register name then, I guess. Those names are confusing... ;-)

phako:
Ok, without the fix, the test was causing nothing to be  shared at all, thus creating a new target accessor all the time. Those checks were simply "delegated" to the base class, which is no longer possible when actually sharing

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
```diff
@@ -330,6 +336,18 @@ namespace ChimeraTK::LNMBackend {
       boost::shared_ptr<LogicalNameMappingBackend>& backend, boost::shared_ptr<NDRegisterAccessor<TargetType>>& target,
       const UndecoratedParams& params) {
     if constexpr(std::is_same_v<TargetType, uint64_t>) {
+      if(params._wordOffsetInRegister != 0) {
+        throw ChimeraTK::logic_error("BitRangePlugin cannot have a word offset: " + params._name);
+      }
+      if(params._numberOfWords > 1) {
+        throw ChimeraTK::logic_error("BitRangePlugin must have size 1, but " + params._name + " has size " +
```
killenb: Actually, it's not  params._name which has the size, but the size is requested by the user when creating the accessor.

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
```diff
@@ -330,6 +336,18 @@ namespace ChimeraTK::LNMBackend {
...
+            std::to_string(params._numberOfWords) + ".");
+      }
+
+      if(params._flags.has(AccessMode::wait_for_new_data) || params._flags.has(AccessMode::raw)) {
```
killenb, line 348: It would be good if the BitRangePlugin would also work for push type accessors.

killenb (approved):

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
phako: There was a reason for not doing that which I currently do not remember

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
phako: Fixed the wordking

backends/LogicalNameMapping/src/LNMBitRangeAccessPlugin.cc
phako: Made a ticket for it

PR #576, opened 2026-08-17
feat/BitRangeInJmap

killenb:
- **fix: large addresses in simpleJson.jmap**
- **feat: bit ranges in NumericAddressedBackend**

### DIFF
--- a/backends/LogicalNameMapping/include/LNMAccessorPlugin.h
+++ b/backends/LogicalNameMapping/include/LNMAccessorPlugin.h
@@ -104,24 +104,12 @@ namespace ChimeraTK::LNMBackend {
   template<typename Derived>
   class AccessorPlugin : public AccessorPluginBase {
    public:
-    /**
-     * The constructor of the plugin should also accept a 3rd argument:
-     *   const std::map<std::string, std::string>& parameters
-     *  Since the parameters are not used in the base class, they do not need to be passed on.
-     */
-    explicit AccessorPlugin(const LNMBackendRegisterInfo& info, size_t pluginIndex, bool shareTargetAccessors = false);
+    explicit AccessorPlugin(const LNMBackendRegisterInfo& info, size_t pluginIndex);
 
    private:
     // we make our destructor private and add Derived as a friend to enforce the correct CRTP
     ~AccessorPlugin() override = default;
     friend Derived;
-
-   protected:
-    /**
-     * Deriving plugins should set this to true if they want to use interlocked access to the same
-     * target accessor. Otherwise different accessors for the same target will given out.
-     */
-    const bool _needSharedTarget;
 
    public:
     /**
@@ -259,7 +247,7 @@ namespace ChimeraTK::LNMBackend {
     uint32_t _shift{0};
     uint32_t _numberOfBits{0};
     bool _writeable{true};
-    uint32_t dataInterpretationFractionalBits{0};
+    int32_t dataInterpretationFractionalBits{0};
     bool dataInterpretationIsSigned{false};
   };
 
@@ -310,9 +298,8 @@ namespace ChimeraTK::LNMBackend {
   /********************************************************************************************************************/
 
   template<typename Derived>
-  AccessorPlugin<Derived>::AccessorPlugin(
-      const LNMBackendRegisterInfo& info, size_t pluginIndex, bool shareTargetAccessors)
-  : AccessorPluginBase(info), _needSharedTarget{shareTargetAccessors}, _pluginIndex(pluginIndex) {
+  AccessorPlugin<Derived>::AccessorPlugin(const LNMBackendRegisterInfo& info, size_t pluginIndex)
+  : AccessorPluginBase(info), _pluginIndex(pluginIndex) {
     FILL_VIRTUAL_FUNCTION_TEMPLATE_VTABLE(getAccessor_impl);
   }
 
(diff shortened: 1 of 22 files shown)


tests/executables_src/testJsonMapFileParser.cpp
```diff
@@ -147,7 +147,7 @@ BOOST_AUTO_TEST_CASE(TestGoodMapFileParse) {
     BOOST_TEST(reg.nElements == 16384);
     BOOST_TEST(reg.elementPitchBits == 64 * 8);
     BOOST_TEST(reg.bar == 13);
-    BOOST_TEST(reg.address == 0x81000000);
+    BOOST_TEST(reg.address == 0x81000);
```
mhier, line 278: What difference does this make? Is it to speed up launch time of the test?

backends/LogicalNameMapping/include/LNMAccessorPlugin.h
```diff
@@ -109,20 +109,13 @@ namespace ChimeraTK::LNMBackend {
      *   const std::map<std::string, std::string>& parameters
      *  Since the parameters are not used in the base class, they do not need to be passed on.
```
mhier: This comment looks stale (already before, but since you touched the arguments  maybe update it)

backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
```diff
@@ -179,8 +179,30 @@ namespace ChimeraTK {
...
+      return false;
+    }
+    if(_wordOffsetInRegister != rhsCasted->_wordOffsetInRegister) {
+      return false;
+    }
+    if(_flags != rhsCasted->_flags) {
+      return false;
+    }
+
+    // Do we have to compare  formula helpers?
```
mhier: Since you are comparing `_info`, I think no. The formula helpers would be different only if the formulas are, which should be part of the `_info` I think?

include/BitRangeAccessorDecorator.h
```diff
@@ -0,0 +1,437 @@
...
+#include <cstdint>
+#include <mutex>
+
+namespace ChimeraTK::detail {
+
+  /********************************************************************************************************************/
+
+  // From https://stackoverflow.com/questions/1392059/algorithm-to-generate-bit-mask
+  constexpr uint64_t getMaskForNBits(uint64_t numberOfBits) {
+    // Prevent warning about undefined behavior if shifting right by 64 bit below
```
mhier:
```suggestion
    // Prevent undefined behaviour if shifting right by 64 bit below
```

include/BitRangeAccessorDecorator.h
```diff
@@ -0,0 +1,437 @@
...
+    _sharedAccessors->removeTransferElement(_target->getId());
+  }
+
+  /********************************************************************************************************************/
+
+  template<typename UserType, bool isRaw>
+  void BitRangeAccessorDecorator<UserType, isRaw>::doPreRead(TransferType type) {
+    _lock.lock();
+
+    // In a transfer group: only swap the shared buffer to the target once, before calling the target's preRead().
```
mhier: This comment and the next line are hard to understand. We always swap the shared buffer into the target, not only in a TransferGroup. If we are in a TransferGroup, we use the counter of the reentrant mutext to check if we are the first accessor with the same target (and hence the same mutex) to execute that step only once. Did I get this correctly?

include/BitRangeAccessorDecorator.h
```diff
@@ -0,0 +1,437 @@
...
+        // To be revisited after fixing https://redmine.msktools.desy.de/issues/12912
+
+        // Modify the bit range in the shared buffer
+        _sharedBuffer->value[0][0] &= ~_maskOnTarget;
+        _sharedBuffer->value[0][0] |= (value << _shift);
+      }
+      else if constexpr(std::is_integral_v<UserType>) {
+        // Raw mode: copy 1:1 from user buffer to shared buffer with no bit manipulation
+        _sharedBuffer->value[0][0] = static_cast<uint64_t>(buffer_2D[0][0]);
+      }
```
mhier, line 336:
```suggestion
      }
      else {
          assert(false);
      }
```
or maybe even a static_assert?

include/BitRangeAccessorDecorator.h
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  /********************************************************************************************************************/
+
+  template<typename UserType, bool isRaw>
+  void BitRangeAccessorDecorator<UserType, isRaw>::doPostRead(TransferType type, bool hasNewData) {
+    auto unlock = cppext::finally([this] { this->_lock.unlock(); });
+
+    // step 1: target postRead() and swap the data into the shared state
+    if(_lock.mutex()->useCount() == _sharedAccessors->instanceCount(_target->getId())) {
```
mhier, line 214: maybe also add some explaining comment that this means the `if` body is executed once for the first accessor (with same target) inside a `TransferGroup`?

include/BitRangeAccessorDecorator.h
```diff
@@ -0,0 +1,437 @@
...
+  }
+
+  /********************************************************************************************************************/
+
+  template<typename UserType, bool isRaw>
+  void BitRangeAccessorDecorator<UserType, isRaw>::replaceTransferElement(
+      boost::shared_ptr<ChimeraTK::TransferElement> newElement) {
+    auto castedToThisType = boost::dynamic_pointer_cast<BitRangeAccessorDecorator<UserType, isRaw>>(newElement);
+
+    // In a transfer group, we are trying to replaced with an accessor. Check if this accessor is for the
```
mhier: that sentence is broken...

mhier (approved):

mhier:
I am still getting an exception when trying to use a bit range register in QtHardMon:
```
Info: An exception was thrown:
BitRangeAccessorDecorator in raw mode requires integral UserType for register '/phasemeter_proto/REG_COUNTER/counter_reset'. Requested type is i.
```
Shouldn’t this work by now?

killenb:
> I am still getting an exception when trying to use a bit range register in QtHardMon:
> 
> ```
> Info: An exception was thrown:
> BitRangeAccessorDecorator in raw mode requires integral UserType for register '/phasemeter_proto/REG_COUNTER/counter_reset'. Requested type is i.
> ```
> 
> Shouldn’t this work by now?

I think it should. The error message seems broken. I suppose it should says "requires unsigned integral user type". And QtHardMon does not look at the raw type in the catalogue but always get (signed) int32. Or the catalogue entry is wrong. I will check.

mhier:
> I think it should. The error message seems broken. I suppose it should says "requires unsigned integral user type". And QtHardMon does not look at the raw type in the catalogue but always get (signed) int32. Or the catalogue entry is wrong. I will check.

Yes, the error message is for sure wrong, it should say "requires uint64 UserType" instead of "integral". Obviously this is not the only problem...

tests/executables_src/testJsonMapFileParser.cpp
killenb: Yes, it's to speed up the tests. The shared memory dummy takes what feels like half a minute, which is very annoying for a test that takes a few seconds, or when using QtHardMon.

backends/LogicalNameMapping/src/LNMBackendVariableAccessor.cc
killenb: I thought the same. I will write this as a comment why it is not part of the comparison.

backends/LogicalNameMapping/include/LNMAccessorPlugin.h
killenb: I don't know what else to write that is not trivial. I'll just remove it.

include/BitRangeAccessorDecorator.h
killenb: Yes, that's correct. I'll update the comment.

include/BitRangeAccessorDecorator.h
killenb: static_assert does not work because the code branches for the other data types are in fact compiled. But they never should be instantiated, so I put the runtime assert.

mhier (approved):

src/DeviceBackendImpl.cc
```diff
@@ -67,4 +69,16 @@ namespace ChimeraTK {
...
 
+  inline boost::shared_ptr<detail::SharedAccessors> DeviceBackendImpl::getSharedAccessors() {
+    std::lock_guard<std::mutex> lk(_sharedAccessorsMutex);
+    if(!_sharedAccessors) {
+      if(!_sharedAccessors) {
+        _sharedAccessors = boost::make_shared<detail::SharedAccessors>();
+      }
+    }
+    return _sharedAccessors;
+  }
```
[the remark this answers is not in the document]
mhier: Incorrect assessment by CoPilot. This is a lazy initialisation. (`inline` keyword is removed now)

PR #577, opened 2026-08-21
Fix: Re-read ISR in handle() to catch interrupts arriving during processing

nshehz:
Changes GenericMuxedInterruptDistributor::handle() to use a do-while loop that re-reads the ISR after clearing, catching interrupts that arrived during distribution of earlier bits. Also moves the clear from per-bit individual clears to a single batch clear after all bits have been distributed. Fixes a race condition observed on sis8300ku hardware where DONE bits arriving during START processing were permanently missed, leaving the ISR stuck.


killenb:
Something with the rebase from the master went completely wrong. I see the content change from the master and the commit messages from the master in the diff. And I don't see your commit.

Re-reading the ISR after clearing without processing the according interrupts does not seem like the correct solution (on first sight). There should be interrupts in the void queue for it, which might not see an active interrupt if it was already cleared. Maybe the void queue is overflowing and we are missing interrupts. I am also not sure whether batch-clearing is a correct, race condition-free method (it might be, did not look at the code). Please review carefully!

nshehz:
After Re-reading they are not merely cleared, but if there is any new interrupt during the processing of queue it is processed.

nshehz:
pull request creation went wrong, I am creating new pull request and shifting comments.

nshehz:
closing. New pull requested created

PR #578, opened 2026-08-21
Fix: Re-read ISR in handle() to catch interrupts arriving during processing

nshehz:
Changes GenericMuxedInterruptDistributor::handle() to use a do-while loop that re-reads the ISR after clearing, catching interrupts that arrived during distribution of earlier bits. Also moves the clear from per-bit individual clears to a single batch clear after all bits have been distributed. Fixes a race condition observed on sis8300ku hardware where DONE bits arriving during START processing were permanently missed, leaving the ISR stuck.


nshehz:
the last pull request creation went wrong so closed that copied the comments here.
copying killenb comments:
Something with the rebase from the master went completely wrong. I see the content change from the master and the commit messages from the master in the diff. And I don't see your commit.

Re-reading the ISR after clearing without processing the according interrupts does not seem like the correct solution (on first sight). There should be interrupts in the void queue for it, which might not see an active interrupt if it was already cleared. Maybe the void queue is overflowing and we are missing interrupts. I am also not sure whether batch-clearing is a correct, race condition-free method (it might be, did not look at the code). Please review carefully!

nshehz:
After Re-reading they are not merely cleared, but if there is any new interrupt during the processing of queue it is processed.

src/async/GenericMuxedInterruptDistributor.cc
```diff
@@ -633,25 +633,32 @@ namespace ChimeraTK::async {
...
+            }
           }
-        }
-      } // for
+        } // for
+        // Clear all processed interrupts at once so to avoid edge-case.
+        clearInterruptsFromMask(ipr);
+        // Re-read ISR to catch any interrupts that arrived during processing above.
+        // These were not in the original ipr and were not cleared.
+        _isr->read();
```
mhier, line 747: duplicate `_isr->read()`, it's already done at the beginning of the loop

nshehz: fixed

nshehz:
TImeout in ISR re-read added.
Watchdog added.

mhier:
There are again commits shown which are already on the master branch. Can you please fix that by executing the following commands (assuming you are on the branch):
```
git pull
git rebase -i origin/master
git push --force-with-lease
```
Thanks.

src/async/GenericMuxedInterruptDistributor.cc
```diff
@@ -534,6 +543,102 @@ namespace ChimeraTK::async {
...
+      std::this_thread::sleep_for(_watchdogInterval);
+
+      if(_stopWatchdog.load()) {
+        break;
+      }
+
+      uint64_t now = _irqRunCount.load(std::memory_order_relaxed);
+
+      // If we have already raised a device exception, stay dormant until the device has genuinely recovered.
+      if(_watchdogAlerted.load()) {
```
mhier, line 563: I think you do not have to track if this thread has already raised an exception. `_backend->isFunctional()` below (Line 593) will be false after `_backend->setException()` has been called, because that internally already sets and checks an atomic flag.

src/async/GenericMuxedInterruptDistributor.cc
```diff
@@ -534,6 +543,102 @@ namespace ChimeraTK::async {
...
+    uint64_t lastCount = _irqRunCount.load(std::memory_order_relaxed);
+
+    while(!_stopWatchdog.load()) {
+      std::this_thread::sleep_for(_watchdogInterval);
+
+      if(_stopWatchdog.load()) {
+        break;
+      }
+
+      uint64_t now = _irqRunCount.load(std::memory_order_relaxed);
```
mhier: Since you are only checking whether the counter has changed or not, a simple boolean flag will also work here. In the `handle()` function below, the flag should be set to true, while here you can simply use`bool runFlag = _irqRunFlag.exchange(false)` to atomically check the state of the flag and reset it.

Also: You do not have to always use `load()` etc. with `std::atomic`. Simple assignments work as well, unless you want to define the memory_order. Here you use `memory_order_relaxed`, which implies no synchronisation being done w.r.t. other threads. Isn't that incorrect since you might concurrently store a `true` value in the other thread which you hence might not see here (immediately)?

src/async/GenericMuxedInterruptDistributor.cc
```diff
@@ -534,6 +543,102 @@ namespace ChimeraTK::async {
...
+          _watchdogAlerted.store(true);
+          continue;
+        }
+
+        // If the ISR is zero the device is simply idle
+        if(pendingInterrupts == 0) {
+          continue;
+        }
+
+        uint64_t episode = _watchdogExceptionCount.fetch_add(1, std::memory_order_relaxed) + 1;
```
mhier: I don't understand this counter. What is an "episode" in this context and why do you need to keep track of it?

src/async/GenericMuxedInterruptDistributor.cc
nshehz: Without the sticky flag, the watchdog re-raises exceptions every time the backend happens to report functional after a reopen, regardless of whether the handler has genuinely recovered.

src/async/GenericMuxedInterruptDistributor.cc
nshehz: It was at least useful for me to know how many time since the last run it has  failed. Removing.

src/async/GenericMuxedInterruptDistributor.cc
nshehz: Thanks. Yes, that works.

src/async/GenericMuxedInterruptDistributor.cc
mhier: All right, understood. Now I see a loop hole (a small one, but still): If the first interrupt after recover never arrives for whatever reason, the watchdog does not notice this because it is still dormant. You could fix that by clearning `_watchdogAlerted` in `activate()` instead of in the `if` condition below.

Fyi: I have considered checking `_asyncDomain->unsafeGetIsActive()` instead, which would allow us to remove the `_watchdogAlerted` flag anyway, but checking that flag requires locking the domain mutex, which isn't nice and may even have a (minor) impact interrupt processing performance.

mhier (approved):

src/async/GenericMuxedInterruptDistributor.cc
nshehz: added.

PR #582, opened 2026-09-03
feat: named channels in NumericAddressedBackend (JMAP)

mhier:


doc/jmapFormat.dox
```diff
@@ -0,0 +1,390 @@
...
+The minimal register entry needs only a \c name and an \c address:
+\verbatim
+{
+  "name": "SomeTopLevelRegister",
+  "address": { "type": "IO", "channel": 0, "offset": 32 }
+}
+\endverbatim
+
+The \c address object has the following members:
+\li \c type - either \c "IO" (the default, a bar on the I/O bus) or \c "DMA". A DMA address is typically used for bulk data acquisition registers.
```
phako: a bar on the I/O bus -> an address on the I/O bus

doc/jmapFormat.dox
```diff
@@ -0,0 +1,390 @@
...
+}
+\endverbatim
+
+\subsection jmap_elements Number of elements and bytes per element
+
+By default a register holds a single element of 4 bytes. A register holding more than one element is declared by giving the number of elements and the size of each element:
+\li \c numberOfElements - the number of elements, defaults to \c 1.
+\li \c bytesPerElement - the size of a single element in bytes, defaults to \c 4.
+
+The elements are placed contiguously in memory:
```
phako: I think that is an implementation detail for the numeric adrressed backend, unsure if that should be in here

doc/jmapFormat.dox
```diff
@@ -0,0 +1,390 @@
...
+    },
+    {
+      "name": "CONTROL",
+      "access": "RW",
+      "address": { "type": "IO", "channel": 2, "offset": "0x8004" }
+    }
+  ]
+}
+\endverbatim
+This yields the registers \c APP/STATUS and \c APP.CONTROL.
```
phako: is the use of two separators intentional here?

phako (approved):

doc/jmapFormat.dox
mhier: Would "address space" fit better instead of "memory"?

doc/jmapFormat.dox
mhier:
```suggestion
\li \c type - either \c "IO" (the default, an address on the I/O bus) or \c "DMA". A DMA address is typically used for bulk data acquisition registers.
```

doc/jmapFormat.dox
mhier: No...

```suggestion
This yields the registers \c APP.STATUS and \c APP.CONTROL.
```

doc/jmapFormat.dox
mhier: I think I will remove the entire sentence.

tests/executables_src/testNumericAddressedBackendRegisterAccessor.cpp
```diff
@@ -653,4 +653,78 @@ BOOST_AUTO_TEST_CASE(testMergeNumericRegistersDifferentTypes) {
 
 /**********************************************************************************************************************/
 
+BOOST_AUTO_TEST_CASE(TestMergeStridedChannelSlices) {
```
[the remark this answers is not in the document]
mhier: Test case names are class names and hence shall be uppercase. All other test case names are incorrect actually.

PR #583, opened 2026-09-07
fix: Support "staged" replacement of shared target

phako:
Situation:

- Accessor A with Target T1 created
- Accessor B with Target T2 created
- T1 is replaced with T2
- Accessor C with Target T3 created
- T2 is replaced with T3

 -> Previously, the assertion would trigger because replacing T2 in B
 would remove it from the map.


phako:
Actually, while writing the commit message, it might be enough to just remove the assert, and have the same end result, but this way it's clearer?

mhier (approved):
Yeah, I don't like to have some intermediate state which is broken and where we have to hope that all possible code paths will clean this up further down again... This looks more cleanly.

PR #584, opened 2026-09-08
fix/feat: named channels and interrupts

mhier:


phako (approved):
I am a bit confused, so we did NOT support interrupted named channels, but were also missing the tests for what we supported?

PR #585, opened 2026-09-09
feat: register unit and description

d-rothe:
* register unit and description can be set in .jmap
* unit and/or description can be overwritten in .xlmap via setDescription plugin
* unit and description are available from accessors and catalogue

maybe it makes sense to first merge Christian's changes and then rebase this on top...


backends/LogicalNameMapping/src/LNMSetDescriptionPlugin.cc
```diff
@@ -0,0 +1,71 @@
...
+    : ChimeraTK::NDRegisterAccessorDecorator<UserType>(target) {
+      if(engineeringUnit) {
+        this->_unit = *engineeringUnit;
+      }
+      if(description) {
+        this->_description = *description;
+      }
+    }
+  };
+
```
mhier: Instead of decorating the accessor, couldn't we overwrite the description and unit when handing out the accessor? Full decoration increases memory consumption (every decorator has it's own data buffer) and slightly the CPU usage (3 additional virtual function calls per read or write transfer). It's not a big deal, but if we can do without, we should IMO.

mhier (approved):
Maybe we should also test that the description and unit is available in the accessor?

backends/LogicalNameMapping/src/LNMSetDescriptionPlugin.cc
d-rothe: It seems not to be a problem to simplify the code in that way - I only have to add setUnit/setDescription to TransferElement

PR #586, opened 2026-09-09
feat(jmap)!: register names are now dictionary keys

mhier:
…e as key

Note: I am creating this PR in place of @vaeng just for technical reasons...

### DIFF
--- a/src/JsonMapFileParser.cc
+++ b/src/JsonMapFileParser.cc
@@ -9,6 +9,8 @@
 
 #include <boost/algorithm/string.hpp>
 
+#include <algorithm>
+#include <map>
 #include <string>
 
 using json = nlohmann::json;
@@ -107,7 +109,6 @@ namespace ChimeraTK::detail {
   /** Representation of an entry in the "addressSpace" section, can be either a register or a module (or both) */
 
   struct JsonAddressSpaceEntry {
-    std::string name;
     std::string engineeringUnit;
     std::string description;
     Access access{Access::accessNotSet};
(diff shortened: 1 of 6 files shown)


vaeng:
@mhier do you still see the channel problem? 
@d-rothe if you think it's okay, please merge, so you can use it to work further

mhier:
> @mhier do you still see the channel problem?

No, I probably simply looked at the wrong branch when I did. Sorry for the confusion.

I would like to suggest a briefer commit message subject:

feat(jmap)!: register names are now dictionary keys

The exclamation mark means that the change is breaking, which is the case if someone already uses jmap files.

src/JsonMapFileParser.cc
```diff
@@ -289,45 +301,53 @@ namespace ChimeraTK::detail {
...
+        fill(my, name, parentName, addressSetByParent);
         my.computeDataDescriptor();
         catalogue.addRegister(my);
         if(!channelTabs.empty()) {
           // create one register entry per named channel of the first channel tab: a read-only 1D slice of the
           // 2D register. The channel's byte offset is folded into the address (so bitOffset == 0), and the full
-          // element pitch is kept as the stride between samples.
-          for(const auto& channel : channelTabs[0].channels) {
-            RegisterPath slicePath = my.pathName / channel.name;
+          // element pitch is kept as the stride between samples. Iterate sorted by byte offset so the created
```
d-rothe: I suggest to unify construction of channelsByOffset here and above in line 245 following, by a small helper function or lambda. That will help us with code maintenance.

d-rothe (approved):

PR #587, opened 2026-09-10
fix: inconsistent handling of relative map file names

mhier:


src/NumericAddressedBackend.cc
```diff
@@ -48,6 +53,36 @@ namespace ChimeraTK {
 
   /********************************************************************************************************************/
 
+  std::string NumericAddressedBackend::resolveMapFileName(const std::string& mapFileName) {
+    // an absolute path is used directly, legacy DMAP entries are already delivered as absolute paths
+    if(mapFileName[0] == '/') {
+      if(boost::filesystem::exists(mapFileName)) {
```
phako: I think we can use std:: here, this is C++17

phako (approved):

src/NumericAddressedBackend.cc
mhier: true, fixed.

backends/SharedDummy/src/SharedDummyBackend.cc
```diff
@@ -140,23 +138,9 @@ namespace ChimeraTK {
...
-    std::string absPathToDmapDir = parserUtilities::convertToAbsolutePath(dmapDir);
-    // the map file is relative to the dmap file location. Convert the relative
-    // mapfilename to an absolute path
-    boost::filesystem::path absPathToMapFile{parserUtilities::concatenatePaths(absPathToDmapDir, mapfileName)};
-    // Possible ./, ../ elements are removed, as the path may be constructed
-    // differently in different client applications
-    return boost::filesystem::canonical(absPathToMapFile).string();
+    // the relative map file path is resolved (relative to the DMAP directory, then to the cwd) in the
+    // NumericAddressedBackend base class
+    return returnInstance<SharedDummyBackend>(address, instanceIdHash, mapFileName, parameters["DataConsistencyKeys"]);
```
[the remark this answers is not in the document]
mhier: That works as intended. For the hash to be identical, we require both an identical CDD as well as the map file to be identical.

PR #588, opened 2026-09-11
fix/feat: double buffered named channels

mhier:


tests/executables_src/testJsonMapFileParser.cpp
```diff
@@ -312,6 +312,52 @@ BOOST_AUTO_TEST_CASE(TestGoodMapFileParse) {
     BOOST_TEST(reg.channels[0].width == 16);
     BOOST_TEST(reg.channels[0].getRawType() == ChimeraTK::DataType("int16"));
   }
+  // A double-buffered named channel slice inherits the parent register's double-buffer configuration. The slice
+  // additionally gets its own BUF0/BUF1 buffer-view registers, identical to the parent's but with the channel byte
+  // offset folded into both buffer addresses.
+  {
+    auto reg = regs.getBackendRegister("DAQ.CTRL.errorI");
+    BOOST_TEST(reg.doubleBuffer.has_value());
+    BOOST_TEST(reg.doubleBuffer->address == 0x40200);
```
phako: Why does the address of the parent point to BUF1 here?

phako (approved):

tests/executables_src/testJsonMapFileParser.cpp
mhier: Actually this is `reg.doubleBuffer->address`, not `reg.address`, so it *should* point to BUF1. On the other hand, it seems we are never really using this address (and neither the plan `reg.address`) in case of double buffering, since we are using the ".BUF0" and ".BUF1" entries instead.

src/JsonMapFileParser.cc
```diff
@@ -351,11 +351,35 @@ namespace ChimeraTK::detail {
...
+              NumericAddressedRegisterInfo sliceBuf0 = slice;
+              sliceBuf0.pathName = slicePath + "/BUF0";
+              sliceBuf0.doubleBuffer.reset();
+              sliceBuf0.registerAccess = NumericAddressedRegisterInfo::Access::READ_ONLY;
+              sliceBuf0.computeDataDescriptor();
+              catalogue.addRegister(sliceBuf0);
+              NumericAddressedRegisterInfo sliceBuf1 = slice;
+              sliceBuf1.pathName = slicePath + "/BUF1";
+              sliceBuf1.doubleBuffer.reset();
+              sliceBuf1.address = my.doubleBuffer->address + channel->offset;
```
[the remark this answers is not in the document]
mhier: That's already broken for 2D registers, even the `doubleBuffer` data structure lacks a field for the bar. I will add a check that the bar has to be identical for both buffers.

PR #589, opened 2026-09-14
test: verify interrupt reads on double-buffered registers

mhier:
NOTE: I am experimenting with an AI agent in this PR. It is currently a bit confusing, since the AI agent posts messages under my user name - this point definitively needs improvement in my agentic work flow. Sorry about any inconvenience and/or confusion this may cause.

## CR-001: Interrupt-driven reads on double-buffered registers (verification)

Add tests proving that interrupt-driven `wait_for_new_data` reads on the full
double-buffered register already work end-to-end in the numeric addressed
backends. The circuit (async `TriggeredPollDistributor` plus
`DoubleBufferAccessor`) is already in place; this change request certifies it
with tests and does not add implementation code.

### Requirements

- The full double-buffered register already supports `AccessMode::wait_for_new_data`
  when the map file declares it interrupt-triggered (`triggeredByInterrupt`).
- On each interrupt, the read must return the buffer that the firmware has
  finished writing (the inactive buffer at the time of the read), exactly as the
  polled `DoubleBufferAccessor` does today.
- The double-buffer firmware handshake (disable swapping, read the current buffer
  number, read the buffer, re-enable swapping) must be performed intact when the
  read is triggered by an interrupt.

### Implementation

- Test-only change request: no production code changes. `NumericAddressedBackend`,
  the async `TriggeredPollDistributor` machinery, `JsonMapFileParser`, and
  `DoubleBufferAccessor` are unchanged.
- The traditional `.map` format does not support double-buffering, so the native
  `DoubleBufferAccessor` path is expressed through a `.jmap` file. Introduces a
  new native register `ASYNC_DBLBUF` (interrupt 1, double-buffer index 2) in
  `tests/simpleJsonFile.jmap`.
- New tests in `tests/executables_src/testDoubleBufferAccessor.cc`:
  - `test_interrupt_driven_read_returns_finished_buffer`: four consecutive buffer
    swaps, every blocking read returns the buffer the firmware just filled.
  - `test_interrupt_driven_read_keeps_handshake_enabled`: the handshake does not
    disable swapping after an interrupt-driven read cycle.
- A fixed BUF0/BUF1 buffer view must not advertise `wait_for_new_data` (out of
  scope); the test only covers the full register.

### Tests

- Full CTest suite built and run; all 6 test cases of `testDoubleBufferAccessor`
  pass, including the two new interrupt-driven tests.
- No regression in the double-buffering/async sub-suite (`testDoubleBufferAccessor`,
  `testDoubleBuffering`, `testAsyncRead`,
  `testAsyncVarAndHierarchicalInterruptsUnified`,
  `testNumericAddressedBackendUnified`, `testJsonMapFileParser`).


tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +94,128 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+// ------------------------------------------------------------
+// Helper: simulate the firmware writing a buffer and issuing the
+// corresponding interrupt. When triggered by an interrupt, the async
+// double-buffer accessor must return the buffer the firmware just
+// finished writing (the inactive buffer at the time of the read).
+//
+// ASYNC_DBLBUF uses double-buffering index 2 (primary buffer at DMA
+// 0x81070, secondary at 0x81270, both mapped to bar 13). The simulated
+// firmware fills one buffer memory, toggles INACTIVE_BUF_ID to point at
+// the other (already idle) buffer, then raises the interrupt.
```
mhier: Make this comment a bit less verbose, e.g. addresses do not have to be mentioned etc.

mhier:
Review feedback addressed: shortened the AsyncDoubleBufferFixture helper comment as requested (dropped the DMA and bar address details, kept the essential explanation of the simulated firmware action). Pushed as fixup commit 9f7376a4. Rebuilt and re-ran the full CTest suite: testDoubleBufferAccessor and the whole double-buffering/async sub-suite pass; the only failures are the three known hardware tests (testPcieBackend, testRegisterAccess, testUioBackendUnified) that need /dev/pcieunidummys6 and /dev/ctkuiodummy and are not runnable here.

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +94,127 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+  AsyncDoubleBufferFixture() {
+    device.open("(dummy?map=simpleJsonFile.jmap)");
+    dummy = boost::dynamic_pointer_cast<DummyBackend>(device.getBackend());
+    BOOST_REQUIRE(dummy);
+
+    accessor = dummy->getRegisterAccessor<uint32_t>("/DAQ/ASYNC_DBLBUF", 1, 0, {AccessMode::wait_for_new_data});
+
+    // Enable double buffering for index 2
+    enable = dummy->getRegisterAccessor<uint32_t>("/DAQ/DOUBLE_BUF/ENA", 1, 2, {});
+    enable->read();
```
mhier: read is not necessary, since we anyway overwrite the data

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +94,127 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+// fills one buffer memory, toggles INACTIVE_BUF_ID to point at the
+// other (already idle) buffer, then raises the interrupt.
+class AsyncDoubleBufferFixture {
+ public:
+  AsyncDoubleBufferFixture() {
+    device.open("(dummy?map=simpleJsonFile.jmap)");
+    dummy = boost::dynamic_pointer_cast<DummyBackend>(device.getBackend());
+    BOOST_REQUIRE(dummy);
+
+    accessor = dummy->getRegisterAccessor<uint32_t>("/DAQ/ASYNC_DBLBUF", 1, 0, {AccessMode::wait_for_new_data});
```
mhier: Avoid using backend API directly here. Either use the `device` for the accessors, or use the `DummyBackend` backdoor accessors. (I think I prefer the second option, backdoor accessors, since they yield in shorter code) This applies to all uses of `getRegisterAccessor` here.

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +94,127 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+// ------------------------------------------------------------
+// Test that an interrupt-triggered wait_for_new_data read of a full
+// double-buffered register returns the buffer the simulated firmware just
+// finished writing, over several consecutive buffer swaps.
+BOOST_AUTO_TEST_CASE(test_interrupt_driven_read_returns_finished_buffer) {
+  AsyncDoubleBufferFixture f;
+  f.device.activateAsyncRead();
+
+  // Initial value arrives while the async domain is activated.
+  f.accessor->readNonBlocking();
```
mhier: Check return value

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +94,127 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+  BOOST_CHECK(f.accessor->dataValidity() != ChimeraTK::DataValidity::faulty);
+
+  // Swap 1: firmware finishes buffer0 (=100) and flips to buffer1.
+  f.firmwareFinishesBuffer(100, 1);
+  f.accessor->read();
```
mhier: Better use `readNonBlocking` and check return value. Also re-run `readNonBlocking` and make sure no more data is available, resp. no data is available before `firmwareFinishesBuffer` is called. This applies to the rest of the test as well.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: The comment in the fixture has been shortened and no longer mentions raw BAR addresses. The buffer memories are now named by register (BUF0/BUF1) and written via DummyBackend backdoor accessors.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Done. The redundant enable->read() has been removed; the enable register is now written directly (enable[0] = 1; enable.write()).

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Done. The accessor under test (which needs wait_for_new_data) is obtained from the device via Device::getOneDRegisterAccessor, as it must exercise the real async path. The ENA and INACTIVE_BUF_ID control registers are also obtained from the device, because the DummyBackend backdoor accessor (DummyRegisterAccessor) cannot handle these 1-bit registers (it requires a 32-bit aligned element pitch). The raw buffer memories (BUF0/BUF1), which are read-only views through the device, are written via DummyRegisterAccessor backdoor accessors as you preferred.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Done. The readNonBlocking() return value is now checked with BOOST_CHECK.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Done. The interrupt-driven tests now use readNonBlocking() throughout with return-value checks (BOOST_CHECK(f.accessor.readNonBlocking())), and verify no data is available before firmwareFinishesBuffer is called and no more data remains after each finished buffer is consumed (BOOST_CHECK(!f.accessor.readNonBlocking())), for both tests and all buffer swaps.

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +93,142 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+
+// ------------------------------------------------------------
+// Helper: simulate the firmware writing a buffer and issuing the
+// corresponding interrupt. When triggered by an interrupt, the async
+// double-buffer accessor must return the buffer the firmware just
+// finished writing (the inactive buffer at the time of the read).
+//
+// ASYNC_DBLBUF uses double-buffering index 2. The simulated firmware
+// fills one buffer memory, toggles INACTIVE_BUF_ID to point at the
+// other (already idle) buffer, then raises the interrupt.
```
mhier: The comments (also those below) violate the 120 character-per-line formatting rule (too short lines). Also the comments are generally too verbose, make them more concise (avoid stating simple facts that are obvious from the code, like the use of DummyBackend backdoor accessors).

doc/change-requests/CR-001-interrupts-with-double-buffering.md
```diff
@@ -0,0 +1,119 @@
...
+
+## Test plan
+
+- New test: async read of a full interrupt+double-buffer register, over several
+  buffer swaps, verifying each returned value corresponds to the buffer the
+  simulated firmware last filled.
+- Full sub-suite run (`ctest`) of the double-buffering and async tests to ensure
+  no regression.
+
+## Tests written (tests-write phase)
```
mhier: Remove these last 3 sections, they do not belong to the original change request and were added during the implementation.

doc/change-requests/CR-001-interrupts-with-double-buffering.md
```diff
@@ -0,0 +1,119 @@
+# CR-001: Interrupt-driven reads on double-buffered registers (verification)
+
+Synopsis: Add tests proving that interrupt-driven `wait_for_new_data` reads on
+the full double-buffered register already work end-to-end in the numeric
+addressed backends. The circuit (async `TriggeredPollDistributor` plus
+`DoubleBufferAccessor`) is already in place; this change request certifies it
+with tests and does not add implementation code.
+
+Status: IN REVIEW
```
mhier: You may now change the status to DONE, merging will be done soon (after review by second person).

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +93,142 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+    inactive[0] = newInactiveBuffer;
+    inactive.write();
+    dummy->triggerInterrupt(1);
+  }
+
+  Device device;
+  boost::shared_ptr<DummyBackend> dummy;
+  OneDRegisterAccessor<uint32_t> accessor;
+  OneDRegisterAccessor<uint32_t> enable;
+  OneDRegisterAccessor<uint32_t> inactive;
```
mhier: Those two should be `DummyRegisterAccessor`s as well, since they are the "backdoor" for the test. Only `accessor` must be a "front-door" regular `OneDRegisterAccessor`.

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +93,142 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+  BOOST_CHECK_EQUAL(f.accessor[0], 400);
+  BOOST_CHECK(!f.accessor.readNonBlocking());
+
+  f.device.close();
+}
+
+// ------------------------------------------------------------
+// Test that the wait_for_new_data read returns newly finished buffer values
+// when the accessor is used with a preceding non-blocking read, and that the
+// double-buffer handshake (enable register stays enabled) is preserved.
```
mhier: This comment is hard to understand and maybe partially stale?

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +93,142 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+BOOST_AUTO_TEST_CASE(test_interrupt_driven_read_keeps_handshake_enabled) {
```
mhier: Isn't this test case already covered by the ordinary double buffering tests? I think we do not need to repeat all tests when testing in combination with interrupts,

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Fixed. The fixture and test comments were condensed to concise, single-line comments (within the 120-column limit) that avoid stating obvious implementation facts (e.g. backdoor accessors). The file is clang-format-clean.

doc/change-requests/CR-001-interrupts-with-double-buffering.md
mhier: Done. Removed the last 3 sections ("Tests written", "Implementation phase result", "Test run result") that were added during implementation and do not belong to the change request.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: I could not make enable/INACTIVE_BUF_ID DummyRegisterAccessor backdoor accessors: DAQ/DOUBLE_BUF/ENA (offset 1234) and DAQ/DOUBLE_BUF/INACTIVE_BUF_ID (offset 1238) are packed 1-bit registers at non-4-byte-aligned addresses, and the DummyBackend DummyRegisterAccessor throws "elementPitchBits/address must be 32-bit aligned" for them (verified by constructing such an accessor). They must therefore stay "front-door" device.getOneDRegisterAccessor accessors, which route through the numeric backends bit-packing. The raw buffer memories BUF0/BUF1 - and everything else - are backdoor DummyRegisterAccessors as you preferred. I added a brief code comment documenting this non-obvious reason.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Fixed. The test header comment was stale/confusing; it now describes the actual purpose: verifying the double-buffer handshake (ENA stays enabled) after an interrupt-driven read.

tests/executables_src/testDoubleBufferAccessor.cc
mhier: Agreed, the value-checking part duplicates the ordinary double-buffer tests. I simplified this test to a single buffer swap and kept only what is unique: that ENA stays enabled (the handshake is not disabled) after an interrupt-driven read, which is a distinct CR-001 requirement.

doc/change-requests/CR-001-interrupts-with-double-buffering.md
mhier: Done, the change request header status is now DONE in the CR-001 document (pushed as part of the fixup).

mhier:
Round-3 review feedback addressed (fixup commit 6937dbab) and pushed:

- Condensed the fixture and test comments to concise single-line comments within the 120-column limit, dropping obvious implementation facts (comment at line 105 and below). File is clang-format-clean.
- Removed the last 3 sections (Tests written / Implementation phase result / Test run result) from the CR-001 document; only the original Requirements / Specifications / Test plan remain.
- Set the CR-001 header status to DONE.
- enable/INACTIVE_BUF_ID stay front-door device accessors: the DummyBackend backdoor accessor throws for these non-32-bit-aligned 1-bit registers (verified); documented this in a brief comment. Everything else uses DummyRegisterAccessor backdoor accessors.
- Rewrote the confusing/stale comment on the handshake test.
- Simplified the handshake test to a single buffer swap, removing the value checks that duplicate the ordinary double-buffer tests and keeping only the unique ENA-stays-enabled check.

Full CTest suite re-run: 70/73 pass. The only 3 failures are the known hardware-gated tests (testPcieBackend, testRegisterAccess, testUioBackendUnified) which need /dev/pcieunidummys6 and /dev/ctkuiodummy and are not runnable in this environment. testDoubleBufferAccessor (all 6 cases, including the two interrupt-driven tests) and the whole double-buffering/async sub-suite pass.

doc/change-requests/CR-001-interrupts-with-double-buffering.md
```diff
@@ -0,0 +1,63 @@
+# CR-001: Interrupt-driven reads on double-buffered registers (verification)
```
vaeng, line 1: This file will be removed upon merge?

mhier: My I idea is to keep it, but we should discuss this maybe tomorrow morning in our meeting with everyone. (I mean fundamentally, whether "change request" documents are wanted or not)

vaeng (approved):

tests/executables_src/testDoubleBufferAccessor.cc
```diff
@@ -93,3 +93,119 @@ BOOST_AUTO_TEST_CASE(test_buffer_selection_current_buffer_1) {
...
+  // No data before the firmware completes the next buffer.
+  BOOST_CHECK(!f.accessor.readNonBlocking());
+
+  f.firmwareFinishesBuffer(500, 0);
+  BOOST_CHECK(f.accessor.readNonBlocking());
+  BOOST_CHECK_EQUAL(f.accessor[0], 500);
+  BOOST_CHECK(!f.accessor.readNonBlocking());
+
+  // The handshake must not disable swapping, so ENA stays enabled.
+  BOOST_CHECK(f.enable[2] == 1);
```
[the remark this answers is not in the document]
mhier: That is sufficiently tested in other tests. We don't need to re-test this in every feature combination.

PR #590, opened 2026-09-16
test: verify interrupt reads on double-buffered named channels

mhier:


tests/executables_src/testNumericAddressedBackendUnified.cpp
```diff
@@ -1203,6 +1203,92 @@ uint32_t DoubleBufferedNamedChannelSlice::_currentBufferNumber = 0;
...
+      for(size_t e = 0; e < nElementsPerChannel(); ++e) {
+        v[0].push_back(buffer1[e]);
+      }
+    }
+    return v;
+  }
+
+  void setRemoteValue() {
+    currentBufferNumber[0] = _currentBufferNumber;
+    _currentBufferNumber = _currentBufferNumber ? 0 : 1; // change current buffer no. 0->1 or 1->0
```
phako: Didn't the linter complain about using implicit converstion to bool here?

tests/executables_src/testNumericAddressedBackendUnified.cpp
```diff
@@ -1203,6 +1203,92 @@ uint32_t DoubleBufferedNamedChannelSlice::_currentBufferNumber = 0;
...
+      }
+    }
+    if(exceptionDummyMuxed->isOpen()) exceptionDummyMuxed->triggerInterrupt(7);
+  }
+
+  void setForceRuntimeError(bool enable, size_t) {
+    exceptionDummyMuxed->throwExceptionRead = enable;
+    exceptionDummyMuxed->throwExceptionWrite = enable;
+    exceptionDummyMuxed->throwExceptionOpen = enable;
+    if(exceptionDummyMuxed->isOpen()) exceptionDummyMuxed->triggerInterrupt(7);
```
phako: I think this is also a linter violation (missing {})

tests/executables_src/testNumericAddressedBackendUnified.cpp
```diff
@@ -1203,6 +1203,92 @@ uint32_t DoubleBufferedNamedChannelSlice::_currentBufferNumber = 0;
...
+      for(size_t e = 0; e < nElementsPerChannel(); ++e) {
+        buffer0[e] = values[0][e];
+      }
+    }
+    else {
+      for(size_t e = 0; e < nElementsPerChannel(); ++e) {
+        buffer1[e] = values[0][e];
+      }
+    }
+    if(exceptionDummyMuxed->isOpen()) exceptionDummyMuxed->triggerInterrupt(7);
```
phako: missing {}

phako (approved):

tests/executables_src/testNumericAddressedBackendUnified.cpp
mhier: Actually no, but I will fix it anyway.

tests/executables_src/testNumericAddressedBackendUnified.cpp
mhier: fixed

tests/executables_src/testNumericAddressedBackendUnified.cpp
mhier: fixed

mhier:
Ok, it's ready to merge, but I want to discuss tomorrow morning in the meeting first if we want to include the change request documents in the repo or not...

mhier:
> **tests/executables_src/testNumericAddressedBackendUnified.cpp:1236**
> 
>     * The generated increment is independent of which buffer is being filled. Since both buffer views start with the same contents, alternating `setRemoteValue()` calls writes the same channel values to buffer 0 and buffer 1, so an async implementation that always reads one buffer can still pass all `CHECK_EQUALITY` checks. Include the selected buffer (or a monotonically changing sequence) in these values so the two buffer paths have distinct expectations.
> 
> 
> ```
>       values[0][e] += uint16_t(37 + 11 * e);
> ```

fixed

PR #591, opened 2026-09-16
feat: Implement Register/Channel muxing parsing in JMAP

phako:


doc/jmapFormat.dox
```diff
@@ -294,9 +290,66 @@ Here \c DMA is a 2D register with two channels, each holding 16384 samples, with
...
+        "representation": { "type": "fixedPoint", "width": 32 }
+      }
+    }
+  }
+}
+\endverbatim
+The \c DAQ.MUX_SEL dot-separated path is normalised to \c /DAQ/MUX_SEL. Channel slices (see \ref jmap_channel_slices) are generated for every channel in the dictionary, regardless of the selector value.
 
-The example above uses a single channel tab, which is the common case. For multi-tab DAQ setups an additional \c tabSelectRegister can be used to switch between the tabs, but the support for multiple tabs is not yet implemented.
+\par Single registers (scalar/1D)
```
mhier: I am wondering if the resulting document structure is good. Isn't this a paragraph below the subsection "Muxed channels (selectedBy)"? That points to "channels", so scalar/1D registers should not be explained below that. Maybe simply changing the title of the subsection is enough, although I feel that the per-channel selector is logically more a special case of the register selector?

tests/executables_src/testJsonMapFileParser.cpp
```diff
@@ -587,6 +675,316 @@ BOOST_AUTO_TEST_CASE(TestInterruptIntegration) {
...
+  BOOST_REQUIRE(reg.channels[1].selectedBy);
+  BOOST_TEST(reg.channels[1].selectedBy->regPath == "/COLLISION/MUX");
+  BOOST_TEST(reg.channels[1].selectedBy->val == 1);
+  BOOST_TEST(!reg.channels[2].selectedBy);
+}
+
+/**********************************************************************************************************************/
+
+// Alternative selectedBy channels that share a byte offset also share the same register address. In a well-formed
+// muxed register the alternatives (different selector values) must resolve to the same address + bit offset.
```
mhier: That ("In a well-formed muxed register the alternatives must resolve to the same address") sounds like a very general restriction. Is this really the case, or is it just pointing out that in this test case with the jmap file at hand this should be the case? We should not restrict the channel layouts between the alternatives to be identical...

tests/executables_src/testJsonMapFileParser.cpp
```diff
@@ -587,6 +675,316 @@ BOOST_AUTO_TEST_CASE(TestInterruptIntegration) {
...
+  BOOST_TEST(acc.isReadOnly());
+
+  dev.close();
+}
+
+/**********************************************************************************************************************/
+
+// Open a slice of the non-muxed SIMPLE2D register; must open without bad_optional_access and
+// without wait_for_new_data.
+BOOST_AUTO_TEST_CASE(TestSelectedByNonMuxedSliceAccessor) {
```
mhier: Isn't that  covered already by the pre-existing named channel slice tests?

Btw: This test case and all test cases which actually obtain accessors should not be in "testJsonMapFileParser.cpp" where we merely want to check the map file parser isolated from the actual backend...

tests/executables_src/testJsonMapFileParser.cpp
```diff
@@ -587,6 +675,316 @@ BOOST_AUTO_TEST_CASE(TestInterruptIntegration) {
...
+
+  // The two muxed alternatives share the same address but differ in their selector value.
+  BOOST_TEST(reg0.address == 1246);
+  BOOST_TEST(reg1.address == 1246);
+  BOOST_TEST(reg0.address == reg1.address);
+}
+
+/**********************************************************************************************************************/
+
+// A muxed channel that also has bit-field sub-entries. The named channel slice collides with the
```
mhier: Does the channel have a bit field child? Or is the selectedBy target a bit field?

(Channels with bit field children do not work yet, I am working on it right now)

mhier (commented):
I feel the tests are not yet well organised. The testJsonMapFileParser test cases should ideally use the MapFileParser in isolation, without a Device or backend.

Maybe the tests which require a broken jmap file to check for an exception should not bring their own file but instead use an existing file and inject the error into that file (parse the file with nlohman::json, change to make it broken and write to some file with a unique name in the build directory)

Also some test cases seem a bit redundant, but maybe I am just confused by the many possibilities.

src/JsonMapFileParser.cc
```diff
@@ -191,42 +191,74 @@ namespace ChimeraTK::detail {
...
+        j = nlohmann::json{{"register", s.regPath}, {"value", s.value}};
+      }
+    };
+
+    struct Channel {
+      std::string engineeringUnit;
+      std::string description;
+      size_t offset;
+      size_t bytesPerElement{4};
+      Representation representation{};
```
mhier: Isn't the linter complaining about the empty `{}`? It's redundant.

tests/executables_src/testJsonMapFileParser.cpp
phako: Yeah for some reason those tests crept in the wrong file and I did not pay proper attention

tests/executables_src/testJsonMapFileParser.cpp
phako: That was meant for this particular case where they share the same parent and do not specify their own offset. Probably not limited to selectedBy and more a general thing, hence redundant.

doc/jmapFormat.dox
phako: I have moved the general documentation to the top to "Register entry", and just added an example and a reference to that in the channel section

tests/executables_src/testJsonMapFileParser.cpp
phako: Tests are removed

src/JsonMapFileParser.cc
phako: Fixed

tests/executables_src/testJsonMapFileParser.cpp
phako: The channels have a bitfield child. I think this test is redundant?

tests/selectedByCases.jmap
```diff
@@ -0,0 +1,159 @@
...
+                "register": "BITFIELD.MUX",
+                "value": 0
+              },
+              "bytesPerElement": 4,
+              "offset": 0,
+              "children": {
+                "Bit0": {
+                  "representation": {
+                    "bitShift": 0,
+                    "width": 1
```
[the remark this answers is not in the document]
phako: That is blatantly wrong, but I removed the whole test anyway

tests/muxedDataAccessor.jmap
```diff
@@ -356,142 +344,139 @@
...
+              "offset": 30,
+              "bytesPerElement": 2,
+              "representation": {
+                "type": "fixedPoint",
+                "width": 16
               }
             }
-          ]
+          }
+        }
```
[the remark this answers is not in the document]
phako: Fixed

src/JsonMapFileParser.cc
```diff
@@ -253,37 +285,41 @@ namespace ChimeraTK::detail {
...
         else if(addressSetByParent) {
-          if(channelTabs.empty()) {
+          if(channels.empty()) {
             if(representation.type == RepresentationType::representationNotSet) {
               throw ChimeraTK::logic_error("Representation not set for register " + parentName / name +
                   " which inherited the address from parent!");
             }
             // If bytesPerElement has not been set in the json file, take it from parent info
             representation.fill(info, 0, (bytesPerElement != 0 ? bytesPerElement : info.elementPitchBits / 8));
+            applyRegisterSelectedBy(info);
```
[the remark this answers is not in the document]
phako: true, this is an oversight in the design.

phako:
I removed a couple of tests that were indeed redundant with other tests (e.g. that wait_for_data is passed through)

src/JsonMapFileParser.cc
```diff
@@ -424,8 +477,8 @@ namespace ChimeraTK::detail {
     }
 
     NLOHMANN_DEFINE_TYPE_INTRUSIVE_WITH_DEFAULT(JsonAddressSpaceEntry, engineeringUnit, description, access,
-        triggeredByInterrupt, numberOfElements, bytesPerElement, address, representation, children, channelTabs,
-        doubleBuffering)
+        triggeredByInterrupt, numberOfElements, bytesPerElement, pitch, address, representation, children, channels,
+        doubleBuffering, selectedBy)
```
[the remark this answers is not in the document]
phako: Tool was updated

phako:
I will add the recursive selectBy in a different PR

mhier (approved):

